PR #4, opened 2020-01-30
Editor config file is added

AlexUstinov:
# Conflicts:
#	Orm.sln

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,263 @@
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
+root = true
+
+[*]
+charset = utf-8
+max_line_length = 120
+
+# Markdown Files
+[*.md]
+trim_trailing_whitespace = false
+
+# C# files
+[*.cs]
+# Indentation options
+# https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference#indent
+
+indent_size = 2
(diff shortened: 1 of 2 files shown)


alex-kulakov:
Thank you for your contribution. In general I agree with everything except a few things.
1) We'd rather prefer to have the option ```csharp_space_after_cast``` set to ```true```, historically we add space after casts
2) According to [official documentation](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-formatting-conventions?view=vs-2019#csharp_space_around_declaration_statements) the option ```csharp_space_around_declaration_statements``` has valid values ```ignore``` and ```false```, not ```do_not_ignore```
3) ```dotnet_naming_rule.underscore_internal_naming.symbols  = underscore_internal``` - we don't use underscore prefix for internal properties, events and fields. You may face some exceptional cases where such prefix exists but in general we avoid it. For example, look at the [FieldAttribute](https://github.com/DataObjects-NET/dataobjects-net/blob/master/Orm/Xtensive.Orm/Orm/Attributes/FieldAttribute.cs) class

AlexUstinov:
I've reworked .editorconfig file.
- All urls are replaced with working ones
- All notes from the comment above are taken into account
- Some suggestions converted to warnings. Please pay attention

alex-kulakov:
Thank you.

PR #5, opened 2020-01-30
NameBuilder concurrency issue fix

AlexUstinov:


alex-kulakov:
Thank you for the contribution. But there are some changes of the part of code you've changed committed to master so it's better to apply your changes to updated class. Tests of caching were also added so you can check your changes with them.

Also notice that instance read-only fields are grouped but your static read-only fields breaks the order. Could you please put static read-only field before group of instance read-only fields and make  the ```fieldNameCache``` be part of the group

AlexUstinov:
PR is updated in according to code review comments

PR #7, opened 2020-02-10
Project build process improvements

AlexUstinov:
Goal of these changes is to make solution structure more lightweight and also make project build easier.
In particular this is achieved through the following
- Two separated solutions for `Weaver` and `Orm` are merged into a single solution;
- Dedicated project for building binaries is no longer needed. It is enough to just run standard `dotnet build` command in solution folder;
- It is no longer required to explicitly reference each file from corresponding `.csproj` file. As a result project files easier to understand;
- All excessive `.cs` files including empty `AssemblyInfo.cs` have been deleted.


alex-kulakov:
I can't build the solution locally after your changes. Lots of errors like "error CS0579: Duplicate 'System.Reflection.AssemblyCompanyAttribute' attribute" with various attributes mentioned

Apart from broken solution I have several questions:
1) It is required ```Xtensive.Orm.Weaver``` project to be built before any project with entities (like projects with tests). This is the reason why we have Build.proj. How do you guarantee that in this single solution?
2) Why an ```obj``` folder appeared in the ```_Build``` directory? Before your changes it contained only ```Debug``` and ```Release``` folders and a ```rev.txt``` file.
3) Somehow when I open the solution in VS 2019 each project contains ```obj``` folder as included. That's weird to me, don't you think? It hasn't happened before.
4) Why AutomobileManagerRole.cs was renamed with AutomobilemanagerRole.cs? It was also committed strangely so git shows only addition of the AutomobilemanagerRole.cs but no removal of the AutomobileManagerRole.cs. That's odd.

AlexUstinov:
Solution is not broken but it sensitive to old `obj` folders inside project directories. Run `git clean -xdf` before first build it will remove old `obj` folders.

1. Build order guarantied by the following reference in `Xtensive.Orm` project
  <ItemGroup Label="Build order">
    <ProjectReference Include="$(SolutionDir)Weaver\Xtensive.Orm.Weaver\Xtensive.Orm.Weaver.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" />
  </ItemGroup>
2. `obj` output is redirected to `_Build` folder to not contaminate project folders with temporary stuff.
3. By default project now includes all files in its folder. Removing old `obj` directories will resolve this issue.
4. This must be an issue I'll take a look.

alex-kulakov:
Thank you, now it builds,

> Build order guarantied by the following reference in Xtensive.Orm project.

I'm not worried then.

The 2nd and 3rd: got it;
Fix of the 4th also helped.

PR #9, opened 2020-02-22
Minor unicode renames and typo fixes

afilatov-st:


alex-kulakov:
Thank you for your contribution.

Everything seems to be OK with changes. But for the future I strongly ask you to write commit messages which reflect what's done. For example, "Removes unicode chars, fixes typos" is a good message and "Minor" is a very bad one. It is bad because it forces people to scan actual changeset of the commit to figure out true purpose of its changes. Good understandable messages save a lot of time. I hope you understand.

PR #10, opened 2020-03-05
WIP: Query traceability

afilatov-st:
This is a preliminary pull request, do not merge it, I just want to get a review.

The idea is to have query traceability. We need a possibility to have a comment at the beginning of each SQL query, specifying the location in code which runs a LINQ translated to that query. We don't want it to affect the performance, so we were thinking to avoid using the Stacktrace class, but instead, use [CallerMemberName] attribute and its friends. Also, we can sacrifice some precision, so it's OK to see a trace of the location where LINQ is declared, not executed.

A new extension method `Trace()` is added, and the proposed syntax is:
`Query.All<Customer>().Trace().Where(c => ...)`

Question:
- Is it possible to avoid using a new type of RSE provider (TraceProvider)? The main reason I've used it because I need a topmost `SELECT` to insert a comment above it. And provider compiler allows accessing a root provider. Is it possible to do the same without TraceProvider?

### DIFF
--- a/Orm/Xtensive.Orm.Tests/Linq/QueryMethodTests.cs
+++ b/Orm/Xtensive.Orm.Tests/Linq/QueryMethodTests.cs
@@ -152,11 +152,15 @@ namespace Xtensive.Orm.Tests.Linq
     [Test]
     public void TraceTest()
     {
-      var subquery = Session.Query.All<Customer>().Trace();
-      var query = Session.Query.All<Customer>().Trace()
+      var subquery = CreateQuery1();
+      var query = CreateQuery2()
         .Where(c => c==Session.Query.Single<Customer>(subquery.FirstOrDefault().Key));
 
       Assert.IsNotEmpty(query.ToArray());
     }
+
+    private IQueryable<Customer> CreateQuery1() => Session.Query.All<Customer>().Trace();
+
+    private IQueryable<Customer> CreateQuery2() => Session.Query.All<Customer>().Trace();
   }
 }
--- a/Orm/Xtensive.Orm.Tests/Linq/QueryMethodTests.cs
+++ b/Orm/Xtensive.Orm.Tests/Linq/QueryMethodTests.cs
@@ -148,5 +148,15 @@ namespace Xtensive.Orm.Tests.Linq
       Assert.That(query, Is.Not.Empty);
       Assert.AreEqual(0, expected.Except(query).Count());
     }
+
+    [Test]
+    public void TraceTest()
+    {
+      var subquery = Session.Query.All<Customer>().Trace();
+      var query = Session.Query.All<Customer>().Trace()
+        .Where(c => c==Session.Query.Single<Customer>(subquery.FirstOrDefault().Key));
+
+      Assert.IsNotEmpty(query.ToArray());
+    }
   }
 }
--- a/Orm/Xtensive.Orm.Tests/Upgrade/RecycledDefinitionTest.cs
+++ b/Orm/Xtensive.Orm.Tests/Upgrade/RecycledDefinitionTest.cs
@@ -210,39 +210,4 @@ namespace Xtensive.Orm.Tests.Upgrade
       return Domain.Build(configuration);
     }
   }
-
-  public static class IQueryableExtensions
-  {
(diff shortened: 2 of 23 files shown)


alex-kulakov:
Hi,
I have a really serious doubt about the concept. You've done it the way when it is possible to put Trace() extension method wherever you want within query. That opens doors for some ridiculous cases when you can put trace between every IQueryable<T> extension and, knowing people, they will put it not one time but several. Several usages per query may also appear in cases when queries are written not by people but constructed from different parts by some third-party APIs.

You either need to test such multiple usages with Joins, Groupings, Selects, EntitySets and so on and guarantee that everything is OK in such multi-usage scenarious or rethink the concept to the point where Trace() is a terminal method after which a person no longer able to continue building the query.

Also it would be better to have trace info in Session events such as QueryExecuting/QueryExecuted and DbCommandExecuting/DbCommandExecuted. This, to my mind, will provide extra flexibility and more use cases. You put trace info directly to SQL query and the only choice to have it is to run a profiling software for RDBMS a person uses, what about getting trace info during debugging or during actual runtime? Events may be helpful. Even you can prepend DbCommand text with this info, or not, it is your choice.

Have you considered something similar? Does something ruined this approach in context of your needs?

afilatov-st:
Hi, thanks for the feedback.

1. About the placement of the `Trace()` method -- I can add a check that it should always be after `All()` method (or similar IQueryProvider methods), otherwise an exception will be thrown during the compilation. Does it sound good?

2. It's not clear why did you mention QueryExecuted and DbCommandExecuted events because when they are fired the query is already executed in SQL Server and the comment becomes useless. Also, it's not clear how the QueryExecuting event is related, because it does not allow to affect the query text generation.
Using DbCommandExecuting instead of translation seems to be an option, but
a) it requires us to change the argument of the events, and it's not clear what should be there? A "tracing context"?
b) I heard an opinion from several former DO developers that proving such a `Trace()` method could make a profit for all other DO users, in this case, the built-in translation would be convenient.

dmitrimaximov:
maybe we can introduce `ITraceable<T>` interface which extends `IQueryable<T>` and will be returned from `Query.All` methods. accordingly, `Trace()` method will be an extension to this interface only.

dmitrimaximov:
in case of events, we can extend the existing events if necessary. the problem we were trying to solve was to deliver the caller information through the entire DataObjects.Net query translation pipeline to SQL.Dom so that it can be translated to a query comment.

afilatov-st:
One thing against the event: The idea with the `Trace()` method looks unobtrusive -- if you don't use it, you will not notice any changes in the API and the query text. However, changing the event means that users will see some argument changes, even if they don't need Trace() functionality.

afilatov-st:
@alex-kulakov can we speed up the discussion? If it's more convenient we can use another channel like Skype.

alex-kulakov:
> About the placement of the Trace() method -- I can add a check that it should always be after All() method (or similar IQueryProvider methods), otherwise an exception will be thrown during the compilation. Does it sound good?

I said that you need to guarantee and prove it by testing. If such think will fail expectedly then it is at least OK.

> It's not clear why did you mention ```QueryExecuted``` and ```DbCommandExecuted``` events because when they are fired the query is already executed in SQL Server and the comment becomes useless. Also, it's not clear how the ```QueryExecuting``` event is related, because it does not allow to affect the query text generation.

I mentioned them because they work in pairs and wrap query translation and execution and they also share args so if you add trace info to ```QueryExecuting``` event arguments it will appear in ```QueryExecuted```, same thing with DbCommandExecuted

> it requires us to change the argument of the events, and it's not clear what should be there? A "tracing context"?

That's question of naming and cannot be the reason against. Just choose the name that fits :)

alex-kulakov:
> in case of events, we can extend the existing events if necessary. the problem we were trying to solve was to deliver the caller information through the entire DataObjects.Net query translation pipeline to SQL.Dom so that it can be translated to a query comment.

Why it is important to have comments in SQL.Dom? Probably I don't understand something. I suppose you have some log of SQL queries and you need to attach caller info to them to analyze this later when you will search for time consuming queries. But why do you need this info though DO query translation pipeline?

afilatov-st:
For visibility: a new PR will be provided with attempt to utilize DbCommandExecuting event.

PR #12, opened 2020-03-17
WIP: Traceability via DbCommandExecuting  event

afilatov-st:
Do not merge, this is a preliminary pull request and it will be closed after feedback received.
This time `SqlSelect` is not touched and instead, `DbCommandExecuting` event is utilized.
Several contexts are extended and they are now passed down to storage driver operations in order to provide TraceInfo for the `DbCommandExecuting` event args.

I want to get feedback about the approach. If it looks good to you, I will do this first in our fork, test on our project, and come back with another pull request in several weeks.

### DIFF
--- a/Orm/Xtensive.Orm/Orm/Tracing/TraceInfo.cs
+++ b/Orm/Xtensive.Orm/Orm/Tracing/TraceInfo.cs
@@ -1,0 +1,16 @@
+namespace Xtensive.Orm.Tracing
+{
+  public class TraceInfo
+  {
+    public string CallerMemberName { get; }
+    public string CallerFilePath { get; }
+    public int CallerLineNumber { get; }
+
+    public TraceInfo(string callerMemberName, string callerFilePath, int callerLineNumber)
+    {
+      CallerMemberName = callerMemberName;
+      CallerFilePath = callerFilePath;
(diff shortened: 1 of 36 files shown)


Orm/Xtensive.Orm/Orm/Tracing/TraceInfo.cs
```diff
@@ -0,0 +1,36 @@
+namespace Xtensive.Orm.Tracing
+{
+  /// <summary>
+  /// Query tracing information gathered when Trace() method is called on a LINQ query.
+  /// </summary>
+  public class TraceInfo
```
dmitrimaximov: maybe introduce class

```
class Caller {
  string MemberName;
  string FilePath;
  int LineNumber;
}
```

just to avoid `Caller` prefix for all the properties?

or maybe the entire `TraceInfo` can be transformed into `CallerInfo` class? Unless we plan to add any other properties to it.

AlexUstinov:
Could you rather introduce an extension point to translation engine?
In this case this tracing capability would be implemented as an extension.
It doesn't look like a solution capable to satisfy everyone's expectations because most people expect to have information about the place in code where query is executed not about a place it is declared.
However even information about declaration point might be crucial in many cases so such an extension capable to provide these data might be very useful.

PR #15, opened 2020-04-02
KeyExpression performance improvements

AlexUstinov:
This pull request includes two major improvements
- When model is locked there is no need to sort key columns while creating `KeyExpression`. Presorted column collection could be taken from model;
- Compiler always creates a closure at very beginning of a method. However in `KeyExpression` class many methods perform a check in the beginning and return pre-calculated value if it exists already. With the aim to avoid unnecessary closure creation calculation part has been moved to a separate methods in such a cases.


alex-kulakov:
Thank you for your contribution. Everything seems to be OK

PR #19, opened 2020-04-07
TypeHelper class code formatting and performance improvements

AlexUstinov:
- Lot of code formatting changes;
- `TypeHelper.Activate` method now uses constructor it found rather than always falling back to `Activator.CreateInstance` method;
- `StringComparison.Ordinal` or `char` arguments are now used while processing strings;
- Performance improvement of `IsNullable` by removing redundant checks;
- Performance improvements of `IsAnonimous` and `IsClosure` methods by faster string comparison also by using `Type.IsDefined` rather than `Attribute.IsDefined` (the faster way to call the same check) and by reordering operations in criteria to put faster operations first;
- Using stored type references rather than `typeof` operators to eliminate expensive system calls


alex-kulakov:
Hello, thank you for the contribution. Seems to be OK

PR #24, opened 2020-04-18
Do not call OrmLog.DebugXxx in case Debug level logging is not enabled

AlexUstinov:
It is too expensive to get values of Strings.Xxx properties when those are not needed for actual logging

### DIFF
--- a/Orm/Xtensive.Orm/Orm/Domain.cs
+++ b/Orm/Xtensive.Orm/Orm/Domain.cs
@@ -7,7 +7,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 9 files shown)


Orm/Xtensive.Orm/Orm/Domain.cs
```diff
@@ -44,6 +45,8 @@ public sealed class Domain : IDisposable, IHasExtensions
     private bool isDisposed;
     private Session singleConnectionOwner;
 
+    private bool IsDebugEventLoggingEnabled { get; set; }
+
```
alex-kulakov, line 49: Do we need a private property that never changes (outside constructor, of course)? I assume ```private readonly``` field would be better for class-wide read-only usage.

AlexUstinov: I've converted this property to readonly field

PR #25, opened 2020-04-18
Convert TranslatorScope class to readonly ref struct

AlexUstinov:
This reduces memory traffic and speedups translation process


alex-kulakov:
Thank you for the contribution. PR is OK.

Issue #27, opened 2020-04-28
For data services only public method may be used in materialization pipeline


AlexUstinov:
In the test case provided below only method with `public` modifier successfully works in materialization pipeline. In other cases query translation error is being generated.
``` csharp
using System;
using System.Collections.Generic;
using System.Linq;
using NUnit.Framework;
using Xtensive.Orm.Configuration;

namespace Xtensive.Orm.Tests.Issues
{
  namespace MaterializationMethodTestsModel
  {
    [HierarchyRoot]
    public class Person : Entity
    {
      public Person(Session session) : base(session) { }

      [Field, Key]
      public int Id { get; private set; }

      [Field]
      public string Name { get; set; }
    }

 }

  public class PersonModel
  {
    public int Id { get; set; }
    public string Name { get; set; }
  }

  public abstract class ServiceBase<TEntity, TModel> where TModel: class, new() where TEntity: class, IEntity
  {
    public IList<TModel> GetAllViaPrivate(Session session) =>
      session.Query.All<TEntity>().Select(i => PrivateSelect(i)).ToList();

    public IList<TModel> GetAllViaPublic(Session session) =>
      session.Query.All<TEntity>().Select(i => PublicSelect(i)).ToList();

    public IList<TModel> GetAllViaProtected(Session session) =>
      session.Query.All<TEntity>().Select(i => ProtectedSelect(i)).ToList();

    public IList<TModel> GetAllViaInternal(Session session) =>
      session.Query.All<TEntity>().Select(i => InternalSelect(i)).ToList();

    public IList<TModel> GetAllViaProtectedInternal(Session session) =>
      session.Query.All<TEntity>().Select(i => ProtectedInternalSelect(i)).ToList();

    private TModel PrivateSelect(TEntity entity) => ProtectedSelect(entity);
    public TModel PublicSelect(TEntity entity) => ProtectedSelect(entity);
    internal TModel InternalSelect(TEntity entity) => ProtectedSelect(entity);
    protected internal TModel ProtectedInternalSelect(TEntity entity) => ProtectedSelect(entity);

    protected abstract TModel ProtectedSelect(TEntity entity);
  }

  public class PersonService : ServiceBase<MaterializationMethodTestsModel.Person, PersonModel>
  {
    protected override PersonModel ProtectedSelect(MaterializationMethodTestsModel.Person entity) =>
      new PersonModel { Id = entity.Id, Name = entity.Name };
  }

[...]

PR #28, opened 2020-04-28
Lookup not only publicly visible methods but private methods as well

AlexUstinov:
This is a suggested fix of #27


alex-kulakov:
Thank you for the fix of the issue #27. I'll merge your changes.

PR #30, opened 2020-05-07
Cache tuple value accessors by argument type.

AlexUstinov:
There is no huge variety of possible value types in any reasonable domain mode.
This amount is restricted by amount of primitive types plus a number of types in certain model.
Taking this into account we can cache accessors.


alex-kulakov:
Thank you for you contribution. The changes are ready to merge so I merge them.

PR #34, opened 2020-06-03
Various fixes

alex-kulakov:
Contains a bunch of small fixes:
1) Fixes issue when ```EntithChangeRegistry.Count``` value becomes corrupted.
2) Fixes reference synchronization during entity removal. Full case represented in ```IssueJira0792_*``` file.
3) Addresses an issue when during constraint validation of a lazy-load field its value happens to be fetched even the field isn't changed
4) Removes extra SELECT wrap for cases when TypeId value is the only calculated column in query. There is no point in it. It also fixes an issue when usage of ```In()``` extension in filter of bulk operations produces invalid query, for instance 
  ```
var idsToUpdate = new[] { 99, 100, 102 };
var updatedCount =  session.Query.All<Bar>()
    .Where(b => b.Id.In(idsToUpdate))
    .Update(bar => new Bar(session) { Name = prefix + bar.Name });
```

### CODE REVIEW

File: Orm/Xtensive.Orm/Orm/Providers/SqlCompiler.Helpers.cs [hunk 1]
```diff
@@ -178,6 +178,16 @@ private static bool IsColumnStub(SqlColumn column)
       return false;
     }
 
+    private static bool IsTypeIdColumn(SqlColumn column)
+    {
+      if (column is SqlUserColumn)
+        return (StringComparer.InvariantCultureIgnoreCase.Compare(column.Name, "TypeId") == 0);
```
AlexUstinov, line 184: Why do you use `StringComparer.InvariantCultureIgnoreCase` comparer? Can't it be the `StringComparer.OrdinalIgnoreCase` comparer or just `String.Equals(column.Name, "TypeId", StringComparison.OrdinalIgnoreCase)`? Otherwise this is going to be an expensive check.

Orm/Xtensive.Orm/Orm/Providers/SqlCompiler.Helpers.cs
```diff
@@ -178,6 +178,16 @@ private static bool IsColumnStub(SqlColumn column)
       return false;
     }
 
+    private static bool IsTypeIdColumn(SqlColumn column)
+    {
+      if (column is SqlUserColumn)
+        return (StringComparer.InvariantCultureIgnoreCase.Compare(column.Name, "TypeId") == 0);
+      var cRef = column as SqlColumnRef;
+      if (!ReferenceEquals(null, cRef))
+        return (StringComparer.InvariantCultureIgnoreCase.Compare(column.Name, "TypeId") == 0);
```
AlexUstinov, line 187: Did you mean `cRef.Name` here?

Orm/Xtensive.Orm/Orm/Internals/ReferentialActions.cs
```diff
@@ -36,7 +36,7 @@ private static void OnClearReference(AssociationInfo association, IEntity owner,
     {
       var nullIsEntity = owner as IHasNullEntity;
       var nullValue = nullIsEntity==null ? null : nullIsEntity.NullEntity;
-      if (nullValue!=null || association.OwnerField.IsNullable)
+      if (nullValue!=null || association.OwnerField.IsNullable || removalContext.Contains((Entity) owner))
```
AlexUstinov, line 39: What if `removalContext` if `null` here?

Orm/Xtensive.Orm/Orm/Providers/SqlCompiler.Helpers.cs
AlexUstinov: Also the same question as above about `InvariantCultureIgnoreCase`

AlexUstinov:
Hi @alex-kulakov 
I've notices UnicodeNamesTest is shown as binary file in github (see 
https://github.com/DataObjects-NET/dataobjects-net/pull/34/files#diff-f36b0aa3eba6c46119b3b079d36a3ce3).
It doesn't show differences because of this.
Could you please check?

alex-kulakov:
Seems to be that it is shoun as binary after one of PRs merged into the master branch where file encoding was changed. I'll fix it in this PR.

alex-kulakov:
I have changed encoding for the file to UTF-8 and has to change it to commit but it still shows it as binary file. Hmmm

Orm/Xtensive.Orm/Orm/Providers/SqlCompiler.Helpers.cs
```diff
@@ -211,6 +221,10 @@ private static bool ShouldUseQueryReference(CompilableProvider origin, SqlProvid
         .Where(i => i >= 0)
         .ToList();
       var containsCalculatedColumns = calculatedColumnIndexes.Count > 0;
+      var typeIdIsOnlyCalculatedColumn = (calculatedColumnIndexes.Count == 1)
+        ? IsTypeIdColumn(sourceSelect.Columns[calculatedColumnIndexes[0]])
+        : false;
+
       var rowNumberIsUsed = calculatedColumnIndexes.Count > 0 && sourceSelect.Columns
         .Select((c, i) => new { c, i })
```
AlexUstinov, line 229:
```suggestion
  var typeIdIsOnlyCalculatedColulmn = calculatedColumnIndexes.Count==1 && IsTypeIdColumn(sourceSelect.Columns[calculatedColumnIndexes[0]]);
```

Orm/Xtensive.Orm/Orm/Providers/SqlCompiler.Helpers.cs
```diff
@@ -211,6 +221,10 @@ private static bool ShouldUseQueryReference(CompilableProvider origin, SqlProvid
         .Where(i => i >= 0)
         .ToList();
       var containsCalculatedColumns = calculatedColumnIndexes.Count > 0;
+      var typeIdIsOnlyCalculatedColumn = (calculatedColumnIndexes.Count == 1)
+        ? IsTypeIdColumn(sourceSelect.Columns[calculatedColumnIndexes[0]])
+        : false;
```
AlexUstinov, line 226:
```suggestion
      var typeIdIsOnlyCalculatedColumn = (calculatedColumnIndexes.Count == 1)
        ? IsTypeIdColumn(sourceSelect.Columns[calculatedColumnIndexes[0]])
        : false;
      var typeIdIsOnlyCalculatedColulmn = calculatedColumnIndexes.Count==1 && IsTypeIdColumn(sourceSelect.Columns[calculatedColumnIndexes[0]]);
```

AlexUstinov: Hi @alex-kulakov. Have you seen this suggestion?

alex-kulakov: Done

Issue #49, opened 2020-07-22
In/Contains within queries have no support for Xtensive.Orm.Key[]


alex-kulakov:
When array of Xtensive.Orm.Key is represented as IEnumerable<Xtensive.Orm.Key> then it works, e.g:
```
Xtensive.Orm.Key[] someKeyArray = GetKeysForQuery();
IEnumerable<Key> keyArrayAsEnumerable = someKeyArray;
var result = session.Query.All<MyEntity>().Where(e => keyArrayAsEnumerable.Contains(e.Key)).ToList();

ICollection<Key> keyArrayAsCollection = someKeyArray;
result = session.Query.All<MyEntity>().Where(e => keyArrayAsCollection.Contains(e.Key)).ToList();
```

But direct usage of the array causes an error, for instance:
```
Xtensive.Orm.Key[] someKeyArray = GetKeysForQuery();
IEnumerable<Key> keyArrayAsEnumerable = someKeyArray;
var result = session.Query.All<MyEntity>().Where(e => someKeyArray.Contains(e.Key)).ToList();
```

In() extension has the same behavior

PR #51, opened 2020-07-22
WIP: DataObjects.Net async API

AlexUstinov:
This PR adds async capabilities to DataObjects.Net.
First of all it makes consistent how regular queries, delayed queries or prefetch queries are being executed asynchronously. Now it is `ExecuteAsync` method in all cases.
Also framework now implelments `IAsyncDisposable` and `IAsyncEnumerable` interfaces where appropriate.

[More detailed explanation is to be done]

PR #52, opened 2020-07-23
Let array be used as collection in queries natively instead of generic inteface representation

alex-kulakov:
Fix for #49 

- Improves item type detection for local collections used with ```In()``` and ```Contains()``` methods within queries.
- Adds variety of new collection usage examples to the original test for the feature.

### DIFF
--- a/Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
+++ b/Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
@@ -1,6 +1,6 @@
-﻿// Copyright (C) 2013 Xtensive LLC.
-// All rights reserved.
-// For conditions of distribution and use, see license.
+// Copyright (C) 2013-2020 Xtensive LLC.
+// This code is distributed under MIT license terms.
+// See the License.txt file in the project root for more information.
 // Created by: Alexey Kulakov
 // Created:    2013.12.30
 
--- a/Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
+++ b/Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
@@ -44,6 +44,25 @@ namespace Xtensive.Orm.Tests.Linq.ContainsOverKeysCollectionTestModel
     [Field]
     public MyEntity2 Entity { get; set; }
   }
+
+  [HierarchyRoot]
+  [KeyGenerator(KeyGeneratorKind.None)]
+  public class ComplexKeyEntity : Entity
+  {
+    [Field, Key(0)]
+    public int Id0 { get; private set; }
+
+    [Field, Key(1)]
+    public int Id1 { get; private set; }
+
+    [Field]
+    public string Name { get; set; }
+
+    public ComplexKeyEntity(int id0, int id1)
+      : base(id0, id1)
+    {
+    }
+  }
 }
 
 namespace Xtensive.Orm.Tests.Linq.ContainsOverKeysCollectionTest
@@ -51,8 +70,22 @@ namespace Xtensive.Orm.Tests.Linq.ContainsOverKeysCollectionTest
   [TestFixture]
   public class ContainsOverKeysCollectionTest : AutoBuildTest
   {
-    private IEnumerable<Key> keysList;
-    private IEnumerable<Key> keysArray; 
+    private IEnumerable<Key> keyListAsEnumerable;
+    private IEnumerable<Key> keyArrayAsEnumerable;
+    private ICollection<Key> keyListAsCollection;
+    private ICollection<Key> keyArrayAsCollection;
+    private List<Key> keyList;
+    private IList<Key> keyListAsInteface;
+    private Key[] keyArray;
+
+    private IEnumerable<Key> complexKeyListAsEnumerable;
+    private IEnumerable<Key> complexKeyArrayAsEnumerable;
+    private ICollection<Key> complexKeyListAsCollection;
+    private ICollection<Key> complexKeyArrayAsCollection;
+    private List<Key> complexKeyList;
+    private IList<Key> complexKeyListAsInteface;
+    private Key[] complexKeyArray;
+
     protected override DomainConfiguration BuildConfiguration()
     {
       var configuration = DomainConfigurationFactory.Create();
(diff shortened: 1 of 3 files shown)


Orm/Xtensive.Orm/Orm/Linq/Translator.Queryable.cs
```diff
@@ -1263,11 +1263,18 @@ private Expression VisitExistsAsInclude(Expression source, LambdaExpression pred
...
       ProjectionExpression visitedSource;
       using (state.CreateScope()) {
-        if (source.IsLocalCollection(context) &&
-            (source.Type.IsGenericType && source.Type.GetGenericArguments()[0].IsAssignableFrom(typeof (Key))) ||
-            (source.Type.IsAssignableFrom(typeof (Key)))) {
-          var localCollecctionKeyType = LocalCollectionKeyTypeExtractor.Extract((BinaryExpression)predicate.Body);
-          state.TypeOfEntityStoredInKey = localCollecctionKeyType;
+        if (source.IsLocalCollection(context)) {
+          var sourceType = source.Type;
+          Type itemType = sourceType.IsGenericType
```
AlexUstinov, line 1268: I think it is better to use `var` for code consistency.

Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
```diff
@@ -44,15 +44,48 @@ public class MyEntity3 : Entity
...
   public class ContainsOverKeysCollectionTest : AutoBuildTest
   {
-    private IEnumerable<Key> keysList;
-    private IEnumerable<Key> keysArray; 
+    private IEnumerable<Key> keyListAsEnumerable;
+    private IEnumerable<Key> keyArrayAsEnumerable;
+    private ICollection<Key> keyListAsCollection;
+    private ICollection<Key> keyArrayAsCollection;
+    private List<Key> keyList;
+    private IList<Key> keyListAsInteface;
```
AlexUstinov, line 78: What about `IReadOnlyCollection` and `IReadOnlyList` interfaces checks?

Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
```diff
@@ -65,139 +98,593 @@ protected override void PopulateData()
...
+
+    private void PopulateKeysCollections(Session session)
+    {
+      keyList = session.Query.All<MyEntity>().AsEnumerable().Take(3).Select(e => e.Key).ToList();
+      keyListAsEnumerable = new List<Key>(keyList);
+      keyListAsCollection = new List<Key>(keyList);
+      keyListAsInteface = new List<Key>(keyList);
+
+      keyArray = session.Query.All<MyEntity>().AsEnumerable().Take(3).Select(e => e.Key).ToArray();
+      keyArrayAsEnumerable = keyArray.AsEnumerable().ToArray();
```
AlexUstinov, line 677: Why it is not enough to use just `AsEnumerable` call?

Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
```diff
@@ -65,139 +98,593 @@ protected override void PopulateData()
...
+      keyArrayAsCollection = keyArray.AsEnumerable().ToArray();
+
+      complexKeyList = session.Query.All<ComplexKeyEntity>().Take(5).Select(e => e.Key).ToList();
+      complexKeyListAsEnumerable = new List<Key>(complexKeyList);
+      complexKeyListAsCollection = new List<Key>(complexKeyList);
+      complexKeyListAsInteface = new List<Key>(complexKeyList);
```
AlexUstinov, line 683: I think it is not good to use `List` or array for all cases. If you are testing an interface like `ICollection` or `IReadOnlyCollection` you implementation must be a custom class implementing only the interface you are going to test. This custom class may be a lightweight wrapper over a `List` or an `Array` but it must be completely unique.
In this approach test can guarantee framework doesn't cast this collection to one of those well known types.
Otherwise if the intention was to use casting to `Array` type for example this must be explicitly mentioned in test code.

Orm/Xtensive.Orm/Orm/Linq/Translator.Queryable.cs
alex-kulakov: Yep. Thank you for mentioning that. There was another problem couple lines down.

Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
alex-kulakov: I needed generic interfaces and types. I took some from the top of my mind. Taking more than one is just my habit. Since  ```typeof (IReadOnlyCollection<Key>).IsGenericType``` and ```typeof (IReadOnlyList<Key>).IsGenericType``` are both true it will work just fine.

Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
alex-kulakov: It is not relevant since you encouraged me to rethink test collections one comment bellow.

Orm/Xtensive.Orm.Tests/Linq/ContainsOverKeysCollectionTest.cs
alex-kulakov: Initially, the test checked how translator detects item type of a given local collection and included only IEnumerable<T> test collections, while working on another problem we've found that it's not possible to use an array without casting it to some generic interface it implements. We decided it is wierd and native usage of array should work too so test was extended with additional representations of native generic type (in this case ```List<T>```) and array. Custom non-generic class implementing one of collection interfaces was forgotten though. I will add it to the test and adjist Item type detection.

To summarize, to me we need following test collections:
- an array used natively (check for array must be separate because Type.IsArray is faster that check for interface implementation);
- a custom generic class implementing ```IEnumerable<T>```, since majority of collections implement it eventually (directly or throug other interfaces from System.Collections.Generic namespace);
- custom non-generic class implementing ```IEnumerable<T>``` to make sure it works fine too.

Orm/Xtensive.Orm/Orm/Linq/Translator.Queryable.cs
```diff
@@ -1,6 +1,6 @@
-// Copyright (C) 2003-2010 Xtensive LLC.
-// All rights reserved.
-// For conditions of distribution and use, see license.
+// Copyright (C) 2010-2020 Xtensive LLC.
```
AlexUstinov, line 1: 2009-2020 ?

Orm/Xtensive.Orm/Orm/Linq/Translator.Queryable.cs
```diff
@@ -1572,5 +1570,11 @@ private Expression VisitContainsNone(Expression setA, Expression setB, bool isRo
         return VisitAll(setB, lambda, isRoot);
       }
     }
+
+    private bool IsKeyCollection(Type localCollectionType)
+    {
+      return (localCollectionType.IsArray && localCollectionType.GetElementType() == typeof(Key))
+        || typeof(IEnumerable<Key>).IsAssignableFrom(localCollectionType);
```
AlexUstinov, line 1577: `typeof()` operator is not free it implies system call.
For the future it is good idea to cache known types into static readonly fields.
Please see https://github.com/DataObjects-NET/dataobjects-net/pull/51/files#diff-216f8b7482a2fe8c2cc2e71cf76eb4a7
or
https://github.com/DataObjects-NET/dataobjects-net/pull/51/files#diff-5c9e5c3c17d64a024a021916060f637d
or
https://github.com/DataObjects-NET/dataobjects-net/pull/51/files#diff-43a94c13d7718af1d89d49a6a40b0374R25

for the reference.
We can discuss this approach.

PS: At the same time `typeof` makes sense for generic type arguments or in the attribute declarations.

alex-kulakov: Yes, I know that but now there is no place for putting such well-known types yet. I'd vote for ```WellKnownTypes``` static class (similar to ```Xtensive.Orm.Linq.WellKnownMembers```) and replace all ```typeof```s across all performance sensitive places. But I feel that it should be in a separate PR.

PR #63, opened 2020-09-14
Async transaction api

AlexUstinov:


Orm/Xtensive.Orm/Orm/Session.Transactions.cs
```diff
@@ -38,6 +39,18 @@ public partial class Session
...
 
+    /// <summary>
+    /// Asynchronously opens a new or joins to the already running transaction.
+    /// </summary>
+    /// <param name="token">The cancellation token to terminate execution if needed.</param>
+    /// <returns>
+    /// A new <see cref="TransactionScope"/> object. Its disposal will lead to either commit
+    /// or rollback of the transaction it controls dependently on <see cref="ICompletableScope.IsCompleted"/> flag.
+    /// </returns>
+    /// <exception cref="InvalidOperationException">There is no current <see cref="Session"/>.</exception>
```
alex-kulakov, line 50: I would add a remark telling that this operation should be awaited before any further operation like it is done for async extensions and ExecuteAsync methods. I won't repeat this phrase for methods below but it needs to be done for all new async public methods in this class within the PR

alex-kulakov (changes requested):
The code is OK. I pointed at the summary of one of the methods introduced in the PR and mentioned need to add a remark that awaiting is required before any further actions. Ideally all ```public async``` methods of public types should have such remark to not have misunderstanding form clients in future.

alex-kulakov:
Will fix it in PR #51

PR #69, opened 2020-09-23
Fix the issue with the multiple In statements in the same query

AlexUstinov:
Saves filter data for different statements with the unique key.

### DIFF
--- a/Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0070_IncludeProviderParameterMismatchWithinParameterContext.cs
+++ b/Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0070_IncludeProviderParameterMismatchWithinParameterContext.cs
@@ -31,41 +31,39 @@ namespace Xtensive.Orm.Tests.Issues
     protected override DomainConfiguration BuildConfiguration()
     {
       var config = base.BuildConfiguration();
-      config.Types.Register(Assembly.GetExecutingAssembly(), typeof (Token).Namespace);
+      config.Types.Register(typeof(Token).Assembly, typeof(Token).Namespace);
       return config;
     }
 
     [Test]
     public void TwoDifferentInStatementsInTheSameQuery()
     {
-      using (var session = Domain.OpenSession()) {
-        using (session.OpenTransaction()) {
-          var tokenId = Guid.NewGuid();
-          var token = new Token(session, tokenId) {
-            Salt = -1
-          };
-          session.SaveChanges();
-          var query = session.Query.All<Token>()
-            .Where(t => t.Salt.In(-1, 1) && t.Id.In(tokenId, Guid.Empty));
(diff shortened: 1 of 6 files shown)


alex-kulakov:
I've attached issue so I'm renaming the test for issue. We will use ```IssueGitHub``` prefix followed by issue number from GitHub, in this case IssueGitHub0070. It is also discovered that the issue affects only ```IncludeProvider/SqlIncludeProvider``` so I'm replacing general ```ParameterMismatch``` with something more specific.

Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0070_IncludeProviderParameterMismatchWithinParameterContext.cs
```diff
@@ -26,7 +26,7 @@ public Token(Session session, Guid id)
 namespace Xtensive.Orm.Tests.Issues
 {
   [TestFixture]
-  public class GitHubIssue_xxx_ParameterMismatch : AutoBuildTest
+  public class IssueGitHub0070_IncludeProviderParameterMismatchWithinParameterContext : AutoBuildTest
```
AlexUstinov, line 33: Maybe it is better to say parameter conflict instead of mismatch?

Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0070_IncludeProviderParameterMismatchWithinParameterContext.cs
```diff
@@ -31,41 +31,39 @@ public class IssueGitHub0070_IncludeProviderParameterMismatchWithinParameterCont
...
+        };
+        session.SaveChanges();
+        var query = session.Query.All<Token>()
+          .Where(t => t.Salt.In(-1, 1) && t.Id.In(tokenId, Guid.Empty));
+        Assert.AreSame(token, query.FirstOrDefault());
       }
     }
 
     [Test]
     public void TwoSimilarInStatementsInTheSameQuery()
```
AlexUstinov, line 59: This test case doesn't show anything different than the first one. Maybe we should just remove it.

alex-kulakov: I would keep the one with different types of parameters it shows the problem better and clearer. I guess you meant the same

Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0070_IncludeProviderParameterMismatchWithinParameterContext.cs
alex-kulakov: Yes, I agree.

alex-kulakov (approved):

PR #80, opened 2020-10-16
WIP: New StorageNode selection mechanism

alex-kulakov:
Instead of opening a session and then select ```StorageNode``` for it we select a node and then open session for the selected node.
It allows asynchronous session opening work fine.

Old method of storage node selection marked ```[Obsolete]``` but works with synchronous session opening.


alex-kulakov:
I'm also going to detect whether old method is applied for the sessions with opened connection and throw an exception telling that this is not supported.

Orm/Xtensive.Orm.Tests/Issues/IssueJira0614_TypeMappingCachingInMaterializationContext.cs
```diff
@@ -375,45 +379,43 @@ public void Test02()
 
       var multinodeDomainConfiguration = CreateConfiguration(typeof(Target.TimesheetCode), DomainUpgradeMode.PerformSafely, "Model2");
       var nodeConfiguration = CreateNodeConfiguration(Node1Name, "Model2", "Model1", DomainUpgradeMode.PerformSafely);
+
       using (var domain = BuildDomain(multinodeDomainConfiguration, nodeConfiguration)) {
-        using (var session = domain.OpenSession()) {
-          session.SelectStorageNode(WellKnown.DefaultNodeId);
+        var selectedNode = domain.SelectStorageNode(WellKnown.DefaultNodeId);
```
AlexUstinov, line 384: Let's instead consider direct usage of `StorageNodeManager` as in the sample below.
```csharp
var storageNode = domain.StorageNodeManager.GetNode(WellKnown.DefaultNodeId);
using (var session = storageNode.OpenSession()) {
  ...
}
```

Orm/Xtensive.Orm.Tests/Issues/IssueJira0614_TypeMappingCachingInMaterializationContext.cs
```diff
@@ -257,12 +263,13 @@ public void Test01()
...
-      var multinodeDomainConfiguration = CreateConfiguration(typeof (Target.TimesheetCode), DomainUpgradeMode.Skip, "Model1");
+      var multinodeDomainConfiguration = CreateConfiguration(typeof(Target.TimesheetCode), DomainUpgradeMode.Skip, "Model1");
       var nodeConfiguration = CreateNodeConfiguration(Node2Name, "Model1", "Model2", DomainUpgradeMode.Skip);
+
       using (var domain = BuildDomain(multinodeDomainConfiguration)) {
-        domain.StorageNodeManager.AddNode(nodeConfiguration);
-        using (var session = domain.OpenSession()) {
-          session.SelectStorageNode(WellKnown.DefaultNodeId);
+        _ = domain.StorageNodeManager.AddNode(nodeConfiguration);
+        var selectedNode = domain.SelectStorageNode(WellKnown.DefaultNodeId);
```
AlexUstinov, line 271: For cases like this, we would introduce `GetOrAddNode` method (see below).
```csharp
var storageNode = domain.StorageNodeManager.GetOrAddNode(nodeConfiguration);
using (var session = storageNode.OpenSession()) {
  ...
}
```

Orm/Xtensive.Orm/Orm/Interfaces/ISelectedStorageNode.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Threading.Tasks;
+using Xtensive.Orm;
+using Xtensive.Orm.Configuration;
+
+namespace Xtensive.Orm.Interfaces
+{
+  /// <summary>
+  /// Defindes <see cref="StorageNode"/>-dependant operations.
+  /// </summary>
+  public interface ISelectedStorageNode
```
AlexUstinov, line 11: I suggest renaming this to 'ISessionSource'.

Orm/Xtensive.Orm/Orm/Interfaces/ISelectedStorageNode.cs
```diff
@@ -0,0 +1,97 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Xtensive.Orm;
+using Xtensive.Orm.Configuration;
+
+namespace Xtensive.Orm.Interfaces
+{
+  /// <summary>
+  /// Defindes <see cref="StorageNode"/>-dependant operations.
```
AlexUstinov, line 9: Did you mean "Defines"?

alex-kulakov: yes

Orm/Xtensive.Orm.Tests/Issues/IssueJira0614_TypeMappingCachingInMaterializationContext.cs
alex-kulakov: Do you build nodes on demand and use them just after build?

Orm/Xtensive.Orm.Tests/Issues/IssueJira0614_TypeMappingCachingInMaterializationContext.cs
alex-kulakov: I thought about it but ```StorageNode``` has no relation to Domain now. ```StorageNode``` is stand-alone item to manage by ```StorageNodeManager```. I didn't want to change it. Plus in my opinion ```Domain``` instance is what people have in, for example, service containers, I would. People are used to it.

Another thing is that ```StorageNode``` is mostly used by us. Not so many people use metadata it stores within. We can transform StorageNode to something people work with on regular basis, like opening sessions, but in this case the workflow should be easy-to-use and clear-to-understand. Opening a session to a StorageNode by the StorageNode looks kind of logical for multi-node ```Domain``` at first glance. Any cons of that?

Orm/Xtensive.Orm.Tests/Issues/IssueJira0614_TypeMappingCachingInMaterializationContext.cs
AlexUstinov: Yes, we do. On application start, we don't have any nodes added. On the first request, we build a domain and then add the corresponding node. All subsequent requests either use the existing node or add a new one on the first request to that node.

Orm/Xtensive.Orm.Tests/Issues/IssueJira0614_TypeMappingCachingInMaterializationContext.cs
AlexUstinov: I don't see any cons.

Orm/Xtensive.Orm.Tests/Issues/IssueJira0614_TypeMappingCachingInMaterializationContext.cs
AlexUstinov: The disadvantage of the suggested approach is that `GetOrAddNode` method takes node configuration as a parameter but the existing `GetNode` method gets just node id. I think we would exclude node id from config and pass it separately. So the sample would transform as follows.
```csharp
var storageNode = domain.StorageNodeManager.GetOrAddNode(nodeId, nodeConfiguration);
using (var session = storageNode.OpenSession()) {
  ...
}
// AddNode method should also be changed for consistency.
_ = domain.StorageNodeManager.AddNode(nodeId, nodeConfiguration);
```

alex-kulakov: NodeConfiguration is required to have nodeId anyway, except for default node. It validates before StorageNode build

AlexUstinov: But it is used only in the validation. The `Validate` method is called by the internal `BuildNode` method. In turn, the `BuildNode` method is called only from the `AddNode`.
As you can see we can easily transform all this chain.

alex-kulakov: That's why I added extra check for ```nodeId``` to ```GetOrAdd```. Take a look at new new commits.

AlexUstinov: Can we keep it this way, or even remove the `GetOrAddNode` method for now?
I'd suggest merging this PR to the 6.0 branch and to the master then.
Once the code is in the master I'd think about further improvements.

alex-kulakov: I will keep StorageNodeManager API the same. GetOrAdd functionality can be implemented as an extension in your solution. Later we could integrate your implementation to DataObjects.Net

AlexUstinov (approved):

PR #83, opened 2020-10-27
Implements async SingleConnection closing on Domain.DisposeAsync()

alex-kulakov:


Orm/Xtensive.Orm/Orm/Domain.cs
```diff
@@ -436,41 +436,49 @@ internal Domain(DomainConfiguration configuration, object upgradeContextCookie,
...
+        else {
+          singleConnectionLocal = SingleConnection;
+          SingleConnection = null;
         }
+      }
 
-        return default;
+      var driver = Handlers.StorageDriver;
+      if (isAsync) {
+        await driver.CloseConnectionAsync(null, singleConnectionLocal).ConfigureAwait(false);
```
AlexUstinov, line 476: `singleConnectionLocal` may be `null` here.

AlexUstinov (changes requested):

Orm/Xtensive.Orm/Orm/Domain.cs
alex-kulakov: How?
First, I prevent concurrent Dispose() callers from going further the first ```lock``` by exclusively setting ``isDisposed``` to ``true```, so the second caller by the time it can enter the first ```lock``` has ```isDisposed==true``` and lives the method. So the rest of Dispose can be executed only for the first caller.
Second, I check for ```SingleConnection```  being null.  If it is then there is no point in going further so I call ```return```. The are two places which set the property's value -```Domain```'s constructor and ```Domain.Dispose()``` method. I assume that the property value can be null in two cases 1) it was initialized with null value on Domain instantiation; 2) I set it by calling ```SingleConnection = null``` in Domain.Dispose() method. The case 1 triggers the check and execution lives the method.

Then, I lock the guard to not allow to change the owner and if there any users of single connection I log warning and leave the method. otherwise I get value of the property to a local variable and set the property's value to null after that. To conclude within lock I either initialize the ```singleConnectionLocal``` variable with non-null value or leave the method with warning logged.

Finally, initialized local variable (otherwise I would get a compilation error because I doesn't set any value on variable declaration) is passed to the driver.

Am I missing something in my thoughts?

AlexUstinov: Sorry, I didn't pay attention to the first lock on `disposeGuard`.

AlexUstinov (approved):

PR #99, opened 2020-11-23
Settle the output paths ambiguity

AlexUstinov:


alex-kulakov:
What is suggested catalog structure? What was wrong with old one?

alex-kulakov:
Let's work out a stable folder structure so we won't have to change it every year or so.

Generally we have two types of build output: binaries and whatever content we distribute (now it is only packages). Maybe get distributions out of binaries? I personally don't want to dig into folders to get packages. In 5.0 we had ```_Build/Binaries``` (bin) for all the binaries and ```_Build/Dist``` for the content we distribute (packages, zip files, etc.) I kind of like it because everything what need to be publishes somewhere is in one place. Neat!

What about the ```tests``` folder? What is the purpose of it? To run tests faster in VS Code or on Mac and Linux?

AlexUstinov:
If we want to standardize project layout we'd better follow the approach described here https://gist.github.com/davidfowl/ed7564297c61fe9ab814/.
The purpose of this PR is to remove the `$(Config)` part from the `BaseIntermediateOutputPath` because otherwise, the project build fails after changing configuration from Release to Debug and vice versa in IDE. It is required to run `dotnet restore` from the command line to fix the issue in the existing structure. `IntermediateOutputPath` now depends on `BaseIntermediateOutputPath` so on a build we don't pollute the `_Build` folder with the empty `obj` folder.
Also, both the `OutputPath` and `PackageOutputPath` are now subfolders of the `BaseOutputPath` what is kind of logical.

The purpose of the `Tests` output separation is the cleaner `lib` folder. It contains only distributable assemblies while tests are executed from another folder.

alex-kulakov (commented):
Please taka a look at the comment. Also some merge conflicts appeared since PR was suggested. Please resolve them.

PR #107, opened 2021-01-18
RedundantColumnRemover no longer removes StoreProvider's columns for groupings

alex-kulakov:
In case of GroupBy there are several SQL queries
1) The main query that gets the groups
2) Additional queries, one per group, that request the group content

```RedundantColumnRemover``` does not take into account  additional queries and  in case of ```Query.Store()``` operation it could remove columns that additional queries can use. (see the test class for query example)

The PR makes column optimization to treat all the columns of ```Query.Store()``` operation as used when grouping is applied. This prevents columns from being removed from Temporary table and additional queries will work as intended.

Additionally, it formats some code and improves ```.ToList()```/```.ToArray()``` operations within ```ColumnMappingInspector``` (base class for ```RedundantColumnRemover```)

### DIFF
--- a/Orm/Xtensive.Orm/Orm/Rse/Transformation/ColumnMappingInspector.cs
+++ b/Orm/Xtensive.Orm/Orm/Rse/Transformation/ColumnMappingInspector.cs
@@ -62,8 +62,8 @@ namespace Xtensive.Orm.Rse.Transformation
       var source = VisitCompilable(provider.Source);
       var sourceMap = mappings[provider.Source];
 
-      var indexColumns = new List<int>();
-      var newMappings = new List<int>();
+      var indexColumns = new List<int>(provider.ColumnIndexes.Count);
+      var newMappings = new List<int>(provider.ColumnIndexes.Count);
 
       var currentItemIndex = 0;
       foreach(var item in provider.ColumnIndexes) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Orm/Xtensive.Orm/Orm/Rse/Transformation/ColumnMappingInspector.cs [hunk 1]
```diff
@@ -31,57 +37,72 @@ public virtual CompilableProvider RemoveRedundantColumns()
...
-        .Where(i => i.First >= 0)
-        .ToList();
-      mappings[provider] = columns.Select(c => c.Second).ToList();
-      var indexColumns = columns.Select(c => c.First).ToList();
-      if (source==provider.Source)
-        return provider;
-      return new SelectProvider(source, indexColumns.ToArray());
+
+      var indexColumns = new List<int>();
+      var newMappings = new List<int>();
```
AlexUstinov, line 66: Shouldn't we specify initial capacity here? It might be equal to the `provider.ColumnIndexes.Count`.

AlexUstinov (dismissed):

Orm/Xtensive.Orm/Orm/Rse/Transformation/ColumnMappingInspector.cs
alex-kulakov: Done

PR #108, opened 2021-01-18
Zombie check before transaction commit or rollback

### LINKED ISSUE #77: Attempt to rollback already closed transaction
There is such behavior in master now:

```csharp
[Test]
public void TransactionTest()
{
    using (var session = Domain.OpenSession()) {
        var outerTx = session.OpenTransaction();

        _ = new Item(session, Guid.NewGuid());
        _ = new Item(session, Guid.NewGuid());
        _ = new Item(session, Guid.NewGuid());
        _ = new Item(session, Guid.NewGuid());

        IEnumerator<Guid> enumerator = null;
        var innerTx = session.OpenTransaction(TransactionOpenMode.New);

        enumerator = session.Query.All<Item>()
          .Select(item => item.Id).AsEnumerable().GetEnumerator();
        _ = enumerator.MoveNext(); // force query to execute and hold the reader
       
         // rollback to savepoint and since there is an open reader
        //  this cause SqlException being thrown.
        innerTx.Dispose();

        // the exception thrown before causes SQL transaction being closed with rollback on server
        // so attempt to rollback it again causes new exception here/
        outerTx.Dispose();
    }
}
```

To me first exception makes sense because if we worked on level of DbTransaction and DbCommand we would get the same error. This exception prevents people from letting enumerator being used outside the region where they suppose to be used.

The second exception can be omitted, thought, due to user has already got the first exception, can't it?

alex-kulakov:
Suggested fix #77

The fix is applied only for SQL Server. The PR checks whether transaction is became zombie before commit or rollback. This happens when some exception appeared and rolled back existing transaction (in this case an attempt to rollback to save point when ```DbReader``` is active).

User will have original exception (e.g. inability to perform rollback to save point due to open reader) instead of overlapping exception from outermost ```TransactionScope.Dispose()```

### DIFF
--- a/Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/Connection.cs
+++ b/Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/Connection.cs
@@ -98,6 +98,74 @@ namespace Xtensive.Sql.Drivers.SqlServer
     }
 
     /// <inheritdoc/>
+    public override void Commit()
+    {
+      EnsureIsNotDisposed();
+      EnsureTransactionIsActive();
+
+      try {
+        if (!IsTransactionZombied()) {
+          ActiveTransaction.Commit();
+        }
+      }
+      finally {
+        ActiveTransaction.Dispose();
+        ClearActiveTransaction();
+      }
+    }
+
+    /// <inheritdoc/>
+    public override async Task CommitAsync(CancellationToken token = default)
+    {
+      EnsureIsNotDisposed();
+      EnsureTransactionIsActive();
+
+      try {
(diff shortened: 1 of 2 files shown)


AlexUstinov (approved):
Is this really actual only for the MSSQL provider?

alex-kulakov:
The #77 issue appears only for MS SQL. I have tried to run the tests on other supported RDBMS.

If we speak about checking connection in general then I couldn't find any error in other RDBMSs to add it as a test to have similar symptoms. Having ```DbTransaction.Connection``` set to ```null``` when connection is corrupted is not general behavior so I think we will fix it later if something shows up for other providers.

PR #109, opened 2021-01-22
ActiveTransactionIsNoLongerUsableException class instead of general InvalidOperationException

SergeiPavlov:
It is inconvenient to catch InvalidOperationException for transaction reprocessing.
We need more specific exception class


SergeiPavlov:
Will be fixed in `ISqlTaskProcessor.ProcessTask()`

PR #111, opened 2021-01-26
Resolves command allocation during corrupted transaction

### LINKED ISSUE #110: SimpleCommandProcessor overrides original exception
The command processor processes ```SqlPersistTask```s by executing them. After each command executed it allocates new command for next operation.

The problem is that command execution can end up with exception (an original exception) which rollback server transaction, after that in  ```finally``` part it tries to allocate new command and triggers another check for transaction being active and the check throws new exception (the exception that overrides the original one).

Need to allocate command for future operations only if current command is successfully executed.

The issue is relevant for both ```6.0``` and ```master``` branches.

alex-kulakov:
Fixes #110

### DIFF
--- a/Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0110_SimpleCommandProcessorOverridesOriginalException.cs
+++ b/Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0110_SimpleCommandProcessorOverridesOriginalException.cs
@@ -1,0 +1,166 @@
+// Copyright (C) 2020 Xtensive LLC.
+// This code is distributed under MIT license terms.
+// See the License.txt file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Transactions;
+using NUnit.Framework;
+using Xtensive.Core;
+using Xtensive.Orm.Configuration;
+using Xtensive.Orm.Model;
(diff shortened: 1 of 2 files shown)


Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0110_SimpleCommandProcessorOverridesOriginalException.cs
```diff
@@ -0,0 +1,166 @@
...
+    [Test]
+    public void MainTest()
+    {
+      var task1State = new OperationState(Domain);
+      var task2State = new OperationState(Domain);
+
+      var task1 = new System.Threading.Tasks.Task(Outer, task1State);
+      var task2 = new System.Threading.Tasks.Task(Outer, task2State);
+      task1.Start();
+      task2.Start();
```
AlexUstinov, line 127: I expect race conditions in this test sometimes. It is even possible task1 is completed before task2 started. 
I think it would be good to add some sort of synchronization here. Something like `ManualResetEvent` would help here.

AlexUstinov (approved):

Orm/Xtensive.Orm.Tests/Issues/IssueGitHub0110_SimpleCommandProcessorOverridesOriginalException.cs
alex-kulakov: Thank you. This is good idea. I applied that.

PR #120, opened 2021-02-15
Postgres translator uses case insensitive ILIKE instead of default LIKE.

NekrasovSt:


alex-kulakov:
We cannot interchange LIKE to ILIKE. This breaks existing behavior for sure. We're not going to do that.

alex-kulakov:
If you need to have ```ILIKE``` it should be optional and provider-specific and it should not break existing behavior. To do this you need to:

1. introduce a descendant of ```SqlLike``` class, for instance, ```SqlILike``` in project for Postgres provider
2. Make a method (or methods) for creating such type of ```SqlExpression``` in ```PostgresqlSqlDom``` class (similar to ```SqlDom.Like()``` methods)
3. Add variative translation for this descendant in the ```Translator``` you have changed ( I assume, just check for ```SqlILike``` type is enough to determine what to use - "LIKE" or "ILIKE")

For ```Linq``` support you need to create a pair of an extension method and its compiler. The extension should work same way on server side (when it is translated to SQL) and on client side (when it is applied to local string values). Take a look at our ```StringExtensions.Like()``` and its compiler in ```StringCompilers```. If you would need to normalize string parameters or compare them please use this [recommendations](https://docs.microsoft.com/ru-ru/dotnet/standard/base-types/best-practices-strings).    Since this is provider specific method it should be placed in the provider which supports it.

Tests are needed as well for queries and for local string values.

If you have questions on how to do certain parts then ask, I will point you to right direction for more specific parts of implementation.

alex-kulakov:
No answer for 25 days so I'm closing the PR

PR #123, opened 2021-03-10
Dedicated Prefetch for DelayedQuery{T}

alex-kulakov:
Helps to avoid session activation in cases like
```csharp
var result = session.Query.CreateDelayedQuery(q => q.All<Authors>().Where(s => s.Id.In(request.AuthorsToFetch)))
    .Prefetch(s => s.Books)
    .ToList();
```

### DIFF
--- a/Orm/Xtensive.Orm/Orm/QueryResult.cs
+++ b/Orm/Xtensive.Orm/Orm/QueryResult.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using JetBrains.Annotations;
 using Xtensive.Orm.Linq.Materialization;
 
 namespace Xtensive.Orm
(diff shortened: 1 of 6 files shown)


AlexUstinov:
Hi
I'd also check the following case.
```csharp
var result = session.Query.Execute(q => q.All<Authors>().Where(s => s.Id.In(request.AuthorsToFetch)))
    .Prefetch(s => s.Books)
    .ToList();
```

alex-kulakov:
I have added following case as well

> Hi
> I'd also check the following case.
> 
> ```cs
> var result = session.Query.Execute(q => q.All<Authors>().Where(s => s.Id.In(request.AuthorsToFetch)))
>     .Prefetch(s => s.Books)
>     .ToList();
> ```

Orm/Xtensive.Orm/Orm/QueryResult.cs
```diff
@@ -29,17 +31,20 @@ public EnumerableReader(IEnumerable<TItem> items)
       }
     }
 
-    private readonly IMaterializingReader<TItem> reader;
     private readonly StateLifetimeToken lifetimeToken;
 
+
+    // DO NOT ENUMERATE this reader anywhere outside this class
+    internal IMaterializingReader<TItem> Reader { get; }
```
AlexUstinov, line 38: I thought it would be a good idea to keep `reader` as a field but add the internal `Session` property instead.

AlexUstinov (approved):

PR #127, opened 2021-04-02
Reprocessing extension without session activations inside

### LINKED ISSUE #126: Reprocessing actions running may get the same session
Since ```AsyncLocal<T>``` propagate its context (and value) to child threads one activated session can be used by more than one thread.

This is why ```Xtensive.Orm.Reprocessing.Tests.Reprocessing.Test()``` fails.

Suggested fix: To not use ```Session``` from ```SessionScope``` and to not activate it before ```ExecutionContext.Function``` invocation. We need either pass a ```Session``` instance from the outside with something like ```IExecuteConfiguration.WithExternalSession(Session)``` or open new every if nothing is passed.

alex-kulakov:
Resolves #126 

**Breaking changes**.

Makes reprocessing not use ```SessionScope``` inside (it does not get session out of it and does not open any activation scopes). To compensate that I provided a way to pass a ```Session``` instance from outside by using ```DomainExtensions.WithExternalSession(this Domain, Session)``` extension or ```IExecutionConfiguration.WithExternalSession(this Domain, Session)```

### DIFF
--- a/Extensions/Xtensive.Orm.Reprocessing/ExecuteActionStrategy.cs
+++ b/Extensions/Xtensive.Orm.Reprocessing/ExecuteActionStrategy.cs
@@ -1,8 +1,7 @@
-﻿using System;
+using System;
 using System.Collections.Concurrent;
 using System.Transactions;
 using Xtensive.Core;
-using Xtensive.Orm.Configuration;
 using SystemTransaction = System.Transactions.Transaction;
 
 namespace Xtensive.Orm.Reprocessing
@@ -56,21 +55,23 @@ namespace Xtensive.Orm.Reprocessing
         try {
           T result;
           var disposeSession = false;
-          //SessionScope sessionScope = null;
           try {
             try {
               var isolationLevel = context.IsolationLevel;
               var currentTransaction = SystemTransaction.Current;
-              if (isolationLevel == IsolationLevel.Unspecified && currentTransaction != null)
+              if (isolationLevel == IsolationLevel.Unspecified && currentTransaction != null) {
                 isolationLevel = currentTransaction.IsolationLevel;
+              }
+
               session = context.ExternalSession;
               if (session == null) {
                 session = context.Domain.OpenSession();
                 disposeSession = true;
-                //sessionScope = session.Activate();
               }
-              if (currentTransaction != null && session.Transaction != null)
+              if (currentTransaction != null && session.Transaction != null) {
                 session.EnsureTransactionIsStarted();
+              }
+
               tran = (currentTransaction != null && currentTransaction.TransactionInformation.DistributedIdentifier != Guid.Empty) ||
                      (session.Transaction != null && session.Transaction.IsDisconnected)
                        ? session.OpenTransaction(isolationLevel)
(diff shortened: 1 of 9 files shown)


Extensions/Xtensive.Orm.Reprocessing/ExecuteActionStrategy.cs
```diff
@@ -47,30 +47,27 @@ public abstract class ExecuteActionStrategy : IExecuteActionStrategy
...
           T result;
-          SessionScope sessionScope = null;
-          try
-          {
-            try
-            {
-              IsolationLevel isolationLevel = context.IsolationLevel;
-              SystemTransaction currentTransaction = SystemTransaction.Current;
+          var disposeSession = false;
+          //SessionScope sessionScope = null;
```
AlexUstinov, line 59: Should we delete this?

Extensions/Xtensive.Orm.Reprocessing/ExecuteActionStrategy.cs
```diff
@@ -47,30 +47,27 @@ public abstract class ExecuteActionStrategy : IExecuteActionStrategy
...
                 isolationLevel = currentTransaction.IsolationLevel;
-              session = SessionScope.CurrentSession;
-              if (session == null)
-              {
+              session = context.ExternalSession;
+              if (session == null) {
                 session = context.Domain.OpenSession();
-                sessionScope = session.Activate();
+                disposeSession = true;
+                //sessionScope = session.Activate();
```
AlexUstinov, line 70: and this

Extensions/Xtensive.Orm.Reprocessing/Internals/ExecuteConfiguration.cs
```diff
@@ -37,14 +33,33 @@ public IExecuteConfiguration WithTransactionOpenMode(TransactionOpenMode transac
       return this;
     }
 
+    public IExecuteConfiguration WithExternalSession(Session session)
```
AlexUstinov, line 36: What do you think about renaming this to just `WithSession` ?

AlexUstinov (approved):

Extensions/Xtensive.Orm.Reprocessing/Internals/ExecuteConfiguration.cs
alex-kulakov: Well, I wanted the name to be as clear as possible. That's why I named it ```.WithExternalSession()```. I believe we can rename WithXXX methods but in ```ExecutionContext``` I'd prefer the name which underlines nature of the instance, maybe keep ```ExternalSession``` or replace it with ```UserSession``` (because we don't open it) or something.

Extensions/Xtensive.Orm.Reprocessing/ExecuteActionStrategy.cs
alex-kulakov: done

Extensions/Xtensive.Orm.Reprocessing/ExecuteActionStrategy.cs
alex-kulakov: done

AlexUstinov (approved):

PR #139, opened 2021-05-21
BooleanExpressionConverter: use Value = 1 instead of Value <> 0

commit-n-run:
Value <> 0 does not utilize index over Value, but Value = 1 does

This change would require rebuilding filtered indices with boolean columns in filter expression.


alex-kulakov:
Hello commit-n-run,

Thank you for contribution. Please apply current formatting rules for the lines you have changed, also update the year in copyright of the file from 2020 to 2021. After that I will be able to merge your changes.

alex-kulakov (approved):

PR #154, opened 2021-07-19
Optimizations: allocations in RecordSetHeader.Join(); use IReadOnlyList instead of ReadOnlyList to save allocations

SergeiPavlov:
Profiling shows significant heap allocations in these places, which are suboptimal.

* Optimizations: allocate List<> with known capacity; use IReadOnlyList instead of ReadOnlyList
* Optimize hot paths: replace LINQ by foreach
* `ReadOnlyList` wrapper is unnecessary, because since .NET 45 `List<>` implements `IReadOnlyList`

### DIFF
--- a/Orm/Xtensive.Orm/Orm/Model/ColumnGroup.cs
+++ b/Orm/Xtensive.Orm/Orm/Model/ColumnGroup.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2003-2021 Xtensive LLC.
+// Copyright (C) 2008-2021 Xtensive LLC.
 // This code is distributed under MIT license terms.
 // See the License.txt file in the project root for more information.
 // Created by: Alexey Kochetov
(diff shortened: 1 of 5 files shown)


alex-kulakov (changes requested):
The changes are mostly OK. Some changes are required though. I have suggested changes to apply where I could. ```SqlColumnCollection``` lacks copyright update and that I couldn't suggest

SergeiPavlov:
> The changes are mostly OK. Some changes are required though. I have suggested changes to apply where I could. `SqlColumnCollection` lacks copyright update and that I couldn't suggest

Changed its copyright line: https://github.com/DataObjects-NET/dataobjects-net/pull/154/commits/b930c310c2df110bd2eeb0e9e1d44c54840c3f03

alex-kulakov (approved):

PR #162, opened 2021-09-22
Better way to handle silent table cleanup on table movement to another hierarchy (see original pull-request)

alex-kulakov:
The original #121 had some drawbacks and appeared to have some issues.

This is an improvement PR which does the job without any additional hits and it is more accurate.

```RecreateTableHint``` no longer usable and marked ```Obsolete```

It also brings more test cases

### CODE REVIEW

**Comment by alex-kulakov** [comment]
@AlexUstinov, a quick flashback for you. Original case was about a table moving to another hierarchy without any table change. That caused silent cleanup of table, because current schema state and target schema state were the same. Such scenario should end up with standard exception telling that there are unsafe changes, when ```PerformSafely``` is used. ```ExtactTableStructureWithGeneratorTest```/  ```ExtactTableStructureNoGeneratorTest``` represent original case for every inheritance schema w/ and w/o using of built-in generators (with and without generator table/sequence).  They contain examples of work  w/ and w/o hints and in both ```Perform``` and ```PerformSafely``` modes.

Then, Sergey Pavlov reported me a bug that showed up after I had merged #121, In his case, he tried to rename a type, but keep the underlying table name the same, so he marked the type with```[TableMapping]``` attribute and added ```RenameTypeHint```, but did not keep the name of Primary Key the same as it was in the tests above, so resulting schema difference had change of primary key (his case is shown in the ```PrimaryKeyNameChangedXXX``` tests). That caused unexpected exception on generation of upgrade actions. Obviously my original solution wasn't good, so I should re-work it.

Other tests make sure that other changes of the upgrading type are handled correctly in pair with original case. For instance, a new field is not an unsafe action by itself, but when underlying table has moved to another type from different hierarchy it should end up with standard error because of unsafe action. Same thing with field type changes, field renames and so on.

**Comment by alex-kulakov** [comment]
Also very important places are in ```HintGenerator```.

1) **```GetRecreatedTypes```.** It tries to find the types that conflict by underlying table - it gets new types and then compares their tables with types that are probably removed (suspicious  types). It compares underlying tables by full table name (which includes database and schema if they suppose to be included). 
**New types collection** is basically ```currentModel.Exclude(extractedModel)``
**Removed types** - ```extractedModel.Exclude(currentModel)```
Removed types can also be suspicious, by default DO tries to resolve suspicious types in UpgradeHintProcessor but types that were failed to resolve remain among suspicious. A great example of suspicious type a an Entity that has changed namespace.

2) **```GenerateCleanupByPrimaryKeyHints```.** There are two stages during upgrade where data cleanup is possible - Before any structural schema changes and after them. DeleteDataHint.PostCopy defines the stage, ```false``` means before, ```true``` means after. Before any work on this problem only data deletions with ```PostCopy = true``` considered to be unsafe. But our case generates data cleanup with ```PostCopy = false```, which is OK.  In most cases this deletions before structural changes are safe, but in our case we  need some of them to be unsafe too. So I added extra info to DeleteDataHint and now further operations can detect unsafe deletions among  ```PostCopy = false``` deletions.

File: Orm/Xtensive.Orm/Orm/Upgrade/Internals/HintGenerator.cs [hunk 1]
```diff
@@ -218,64 +219,73 @@ private void GenerateCopyColumnHint(CopyFieldHint hint)
...
+              var identities1 = new[] { new IdentityPair(
+                GetColumnPath(type, GetTypeIdMappingName(type)),
+                removedType.TypeId.ToString(),
+                true)
+              };
+              schemaHints.Add(
+                new DeleteDataHint(GetTablePath(type), identities1, isMovedToAnotherHierarchy, conflictsWithNewType));
+            }
+          }
+          else if (isMovedToAnotherHierarchy) {
```
AlexUstinov, line 250: Why this condition is different from the same one in the other two cases?

alex-kulakov: this is mistake

Orm/Xtensive.Orm/Modelling/Comparison/Hints/DeleteDataHint.cs
```diff
@@ -17,20 +17,46 @@ namespace Xtensive.Modelling.Comparison.Hints
...
     /// these records are still necessary during upgrade to be copied, but must be removed on its
     /// completion.
     /// </summary>
-    public bool PostCopy { get; private set; }
+    public bool PostCopy => info.HasFlag(DeleteDataHintInfo.PostCopy);
+
+    /// <summary>
+    /// Gets a value indicating whether cause of data deletion is due to table have changed its owner type.
+    /// </summary>
+    public bool DueToTableOwnerChange => info.HasFlag(DeleteDataHintInfo.TableMovement);
```
AlexUstinov, line 42: What is the resulting IL code? I suspect the `HasFlag` call may cause boxing.
Instead, we can have just a bitwise operation
```csharp
public bool DueToTableOwnerChange => (info & DeleteDataHintInfo.TableMovement) > 0;
```

Orm/Xtensive.Orm/Modelling/Comparison/Hints/DeleteDataHint.cs
```diff
@@ -17,20 +17,46 @@ namespace Xtensive.Modelling.Comparison.Hints
...
+
     /// <summary>
-    /// Gets or sets a value indicating whether deletion must be performed after completion of copy data hint processing.
+    /// Gets a value indicating whether deletion must be performed after completion of copy data hint processing.
     /// Normally this flag is used to remove records related to types moved to other hierarchies -
     /// these records are still necessary during upgrade to be copied, but must be removed on its
     /// completion.
     /// </summary>
-    public bool PostCopy { get; private set; }
+    public bool PostCopy => info.HasFlag(DeleteDataHintInfo.PostCopy);
```
AlexUstinov, line 37: Can we rename this one to eg `IsPostCopyCleanup`?

Orm/Xtensive.Orm/Modelling/Comparison/Hints/DeleteDataHint.cs
AlexUstinov: The name would be `IsOwnerChangeCleanup`

Orm/Xtensive.Orm/Modelling/Comparison/Hints/DeleteDataHint.cs
```diff
@@ -17,20 +17,46 @@ namespace Xtensive.Modelling.Comparison.Hints
   [Serializable]
   public class DeleteDataHint : DataHint
   {
+    [Flags]
+    internal enum DeleteDataHintInfo : byte
```
AlexUstinov, line 21: `CleanupReason`

Orm/Xtensive.Orm/Modelling/Comparison/Hints/DeleteDataHint.cs
```diff
@@ -17,20 +17,46 @@ namespace Xtensive.Modelling.Comparison.Hints
...
+    {
+      None = 0,
+      PostCopy = 1,
+      TableMovement = 2,
+      All = PostCopy | TableMovement
```
AlexUstinov, line 26: Is there any meaningful name for this aggregate value? Can we just remove it?

Orm/Xtensive.Orm/Modelling/Comparison/Hints/DeleteDataHint.cs
alex-kulakov: > What is the resulting IL code? I suspect the `HasFlag` call may cause boxing. Instead, we can have just a bitwise operation
> 
> ```cs
> public bool DueToTableOwnerChange => (info & DeleteDataHintInfo.TableMovement) > 0;
> ```

Yes, it is. I'll change it

Orm/Xtensive.Orm/Orm/Upgrade/Internals/HintGenerator.cs
```diff
@@ -218,83 +225,151 @@ private void GenerateCopyColumnHint(CopyFieldHint hint)
...
     {
       var removedTypeAndAncestors = new HashSet<StoredTypeInfo>(removedType.AllAncestors.Length + 1);
       removedType.AllAncestors.Append(removedType).ForEach(t => removedTypeAndAncestors.Add(t));
 
+
+      var descendantsToHash = (cleanupInfo & CleanupInfo.ConflictByTable) != 0
+        ? removedType.AllDescendants
+        : Array.Empty<StoredTypeInfo>();
+      var descendants = new HashSet<StoredTypeInfo>(descendantsToHash.Length);
+      descendantsToHash.ForEach(t => descendants.Add(t));
```
AlexUstinov, line 360: `descendants.UnionWith(descendantsToHash);`

AlexUstinov (approved):

PR #191, opened 2021-11-12
Avoid unnecessary array copy before calling .Remap(); new() -> ValueTuple; 

SergeiPavlov:
* Refactor `.Equals()` methods
* Change `Select(..., params int[])` arguments type -> `Select(..., IReadOnlyList<int>)` to avoid conversions

### DIFF
--- a/Orm/Xtensive.Orm.Tests/Rse/HeaderParseTest.cs
+++ b/Orm/Xtensive.Orm.Tests/Rse/HeaderParseTest.cs
@@ -54,7 +54,7 @@ namespace Xtensive.Orm.Tests.Rse
           ResetState(state);
 
           // Select Id, TypeId, Title
-          CompilableProvider rsTitle = rsMain.Select(0, 1, 2);
+          CompilableProvider rsTitle = rsMain.Select(new[] { 0, 1, 2 });
           UpdateCache(session, rsTitle.GetRecordSetReader(session, parameterContext));
           state = Session.Current.EntityStateCache[key, true];
           Assert.IsNotNull(state);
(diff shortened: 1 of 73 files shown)


alex-kulakov:
@SergeiPavlov, I see that ColumnGatherer.GetColumns() changed return type to IEnumerable<T> which forced a lot of places to make .ToList() or ToArray() but in other places. Did it worth it? What scenario overweight these enumerations + possible number of Array.Copy? Please, point me.

SergeiPavlov:
> @SergeiPavlov, I see that ColumnGatherer.GetColumns() changed return type to IEnumerable which forced a lot of places to make .ToList() or ToArray() but in other places. Did it worth it? What scenario overweight these enumerations + possible number of Array.Copy? Please, point me.

There are a few places whe we nneed not call neither `.ToList()` nor `.ToArray()`, but just to enumerate:
```
          var columns = orderByProjector
            .GetColumns(ColumnExtractionModes.TreatEntityAsKey | ColumnExtractionModes.Distinct);
          foreach (var c in columns) {
            if (!sortColumns.ContainsKey(c)) {
              sortColumns.Add(c, direction);
            }
          }
```

```
      var keySegment = visitedSource.ItemProjector.GetColumns(ColumnExtractionModes.TreatEntityAsKey);
      var keyPairs = keySegment
        .Select((leftIndex, rightIndex) => new Pair<int>(leftIndex, rightIndex))
        .ToArray();
```

alex-kulakov:
@SergeiPavlov If we had both results would it be better?

alex-kulakov (approved):

PR #195, opened 2021-11-24
Session.Tag() method with scoped Tagging

SergeiPavlov:
Allows to tag  `.SingleX()`  & `.Prefetch()`  queries


alex-kulakov:
I think session level tagging should work not only for two operations but for almost any of queries. This is clearly not ready for master. I'm switching the target branch.

SergeiPavlov:
It works for all operations. But for  queries returning `IQueryable<>` we alreadfy had `.Tag()` extension

alex-kulakov:
No, it doesn't work for all operations. To me ```.Tag``` extension is per-query solution, session level should work for all queries within tag region. Why should anybody tag session and then be forced to tag queries anyway? This is not a user-friendly approach. Per-query tagging should be additional to session-wide tags.

The cases I see for session tagging:
- case 1 - tag all queries  within session;
- case 2 - tag all queries within session + additional tags for some particular queries;
- case 2 - tag group of queries within session;
- case 3 - tag group of queries within session + additional tags for some particular queries.

I'm pretty sure that it would be beneficial even for you in future and there wouldn't be a need to return and redo current implementation to more convenient one.

alex-kulakov:
Also we have a lot of scopes, and this is kind of known thing in DO, so I would rename ```TagContext``` to ```TagScope``` to be more consistent with naming, sounds reasonable?

alex-kulakov:
Not all Prefetch operations work properly. Prefetch of an EntitySet field does not mark prefetch queries with tags.

Here's the test

```csharp
    [Test]
    public void SessionTagPrefetchEntitySetTest()
    {
      var allCommands = new List<string>();

      using (var session = Domain.OpenSession())
      using (var tx = session.OpenTransaction()) {
        var book = new Author();
        _ = book.Books.Add(new Book() { Name = "Prefetch EntitySet Book #1" });
        _ = book.Books.Add(new Book() { Name = "Prefetch EntitySet Book #2" });

        book = new Author();
        _ = book.Books.Add(new Book() { Name = "Prefetch EntitySet Book #3" });
        _ = book.Books.Add(new Book() { Name = "Prefetch EntitySet Book #4" });

        tx.Complete();
      }

      using (var session = Domain.OpenSession())
      using (var outermost = session.Tag("outermost"))
      using (var tx = session.OpenTransaction()) {

        session.Events.DbCommandExecuting += SqlCapturer;
        _ = session.Query.All<Author>().Prefetch(a => a.Books).ToArray();
        session.Events.DbCommandExecuting += SqlCapturer;

        Assert.That(allCommands.Count, Is.EqualTo(2));
        Assert.IsTrue(allCommands[1].Contains("/*outermost*/"));

        allCommands.Clear();
      }

      void SqlCapturer(object sender, DbCommandEventArgs args)
      {
        allCommands.Add(args.Command.CommandText);
      }
    }
```

SergeiPavlov:
renamed `TagContext` -> `TagScope`

SergeiPavlov:
> Not all Prefetch operations work properly. Prefetch of an EntitySet field does not mark prefetch queries with tags.

Fixed tagging Prefetches

alex-kulakov:
Is it correct that we have two identical tags in case of joins or set operations?

What you've done to ```.All()``` is similar to this:

```csharp
using (var session = Domain.OpenSession())
using (var tx = session.OpenTransaction()) {
  var query = session.Query.All<SomeEntity>();
  foreach (var tag in session.Tags) {
    query = query.Tag(tag);
  }
  query = query.Where(e=>e.Id < 1000);

  var anotherQuery = session.Query.All<SomeEntity>();
  foreach (var tag in session.Tags) {
    anotherQuery = anotherQuery.Tag(tag);
  }
  anotherQuery = anotherQuery.Where(e=>e.Id > 2000 && e.Id < 3000);

  // here we will get all session tags doubled
  // what if there will be several joins and unions. tags could repeat x3, x5, x10 times
  var finalQueryResult = query.Union(anotherQuery).ToArray();
}
```

Do you need a consultation?

SergeiPavlov:
> Do you need a consultation?

You are right. I don't see easy ways to avoid it, except adding some logic into `.Union()` to excude doubled tags

alex-kulakov:
As I see it, our goal is to have functionally something like this 

```csharp
using (var session = Domain.OpenSession())
using (var tx = session.OpenTransaction()) {
  var query = session.Query.All<SomeEntity>();
  query = query.Where(e=>e.Id < 1000);

  var anotherQuery = session.Query.All<SomeEntity>();
  anotherQuery = anotherQuery.Where(e=>e.Id > 2000 && e.Id < 3000);

  var finalQuery = query.Union(anotherQuery);
  foreach (var tag in session.Tags) {
    finalQuery = finalQuery.Tag(tag);
  }
}

So tags are applied not when query construction happens but when it is being translated to be executed.

Two options come to my mind:
- query expression modification;
- providers modification;

Both of them have caviars, for example tail expression can be scalar like "session.Query.All<T>().FirstOrDefault()". I need to think of both options.

Adding ```Tag()``` extension call as tail expression would automatically be supported by current Translator but I would think of edge cases when it might not work.

Wrapping result Providers tree looks easier at least for now. I would look at the result Providers tree of the query above and tried to reproduce it. A good place where you can look at the Providers tree is ```CompilationServices.Compile()``` method, I don't know whether you're familiar with this part of code.  ```TranslatorContext``` has access to session instance in constructor so  tags could be captured here to be applied.

Maybe I will come up with something more clever tomorrow. I'll tell you if I do.

**Comment by SergeiPavlov** [comment]
Moved tagging into PreCompiler. It solves problem with .Union()

**Comment by alex-kulakov** [comment]
I see double-tagging in EntitySet fields prefetch. There is a test for this case and it fails. Probably you skipped this

**Comment by alex-kulakov** [comment]
For test purposes I added some tag to all sessions opened synchronously across the ```Xtenisive.Orm.Tests``` - a lot of tests are failed.

**Comment by SergeiPavlov** [comment]
> I see double-tagging in `EntitySet` fields prefetch. There is a test for this case and it fails. Probably you skipped this. You need to remove the code you added previously from `EntitySetTask` as you did with `EntityGroupTask`.

Done

**Comment by alex-kulakov** [comment]
I suggest following implementation
First, change ```Xtensive.Orm.Linq.TranslatorContext``` and add ```SessionTags``` property with following initialization in ```ctor```
```csharp
public TranslatorContext(Session session, CompilerConfiguration rseCompilerConfiguration, Expression query,
    CompiledQueryProcessingScope compiledQueryScope)
{
  ArgumentValidator.EnsureArgumentNotNull(session, nameof(session));
  ArgumentValidator.EnsureArgumentNotNull(rseCompilerConfiguration, nameof(rseCompilerConfiguration));
  ArgumentValidator.EnsureArgumentNotNull(query, nameof(query));

  Domain = session.Domain;
  RseCompilerConfiguration = rseCompilerConfiguration;
  // this allows to completely turn off tagging by option in domain config
  // while all code remains the same
  SessionTags = (Domain.Configuration.TagsLocation != TagsLocation.Nowhere)
    ? session.Tags
    : null;

  // Applying query preprocessors
  query = Domain.Handler.QueryPreprocessors
    .Aggregate(query, (current, preprocessor) => ApplyPreprocessor(preprocessor, session, current));

  //....
}
```
After this in ```Translator.Materialization.cs``` in method Translate(ProjectionExpression, IEnumerable<Parameter<Tuple>>) add tags to the end of current projection like so
```csharp
  internal TranslatedQuery Translate(ProjectionExpression projection,
      IEnumerable<Parameter<Tuple>> tupleParameterBindings)
    {
      var result = projection;
      if (context.SessionTags != null) {
        result = ApplySessionTags(projection, context.SessionTags);
      var newItemProjector = projection.ItemProjector.EnsureEntityIsJoined();
      result = projection.Apply(newItemProjector);

      var optimized = Optimize(result);

      // Prepare cached query, if required
      var prepared = compiledQueryScope != null
        ? PrepareCachedQuery(optimized, compiledQueryScope)
        : optimized;
      //....
}
```

alex-kulakov:
What about tags in SessionConfiguration? Would it be useful? It could be implemented based on current API.

SergeiPavlov:
> What about tags in SessionConfiguration? Would it be useful? It could be implemented based on current API. @LookBehind, you also were interested in tagging. What do you think?

Lets ask also @AlexUstinov

SergeiPavlov:
>...
> Such implementation seems to not break queries. Also pre-compilers order as existance might be changed by overriding `CreatePreCompiler` method, which is not good.

Thank may be, but how to implement `ApplySessionTags(ProjectionExpression, IReadOnlyList<string> tags)` ?

alex-kulakov:
> Thank may be working, but how to implement ApplySessionTags(ProjectionExpression, IReadOnlyList<string> tags) ?

Oh, yes, I forgot to post it
```csharp
    private static ProjectionExpression ApplySessionTags(ProjectionExpression origin, IReadOnlyList<string> tags)
    {
      var currentProjection = origin;
      foreach (var tag in tags) {
        var projector = currentProjection.ItemProjector;
        var newDataSource = currentProjection.ItemProjector.DataSource.Tag(tag);
        var newItemProjector = new ItemProjectorExpression(projector.Item, newDataSource, projector.Context);
        currentProjection = new ProjectionExpression(currentProjection.Type, newItemProjector,
          currentProjection.TupleParameterBindings, currentProjection.ResultAccessMethod);
      }
      return currentProjection;
    }
```

alex-kulakov:
Now it is ready to merge to master directly. I'm switching back

alex-kulakov (approved):

PR #198, opened 2021-12-02
Change DATEDIFF to DATEDIFF_BIG for Sql Server v13+

alex-kulakov:
DATEDIFF_BIG has bigger resolution and does not throw overflow exception even on vastly spaced dates. This allows to get rid of compute-expensive formula that served overflow prevention (as I understood).

### DIFF
--- a/Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v09/Compiler.cs
+++ b/Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v09/Compiler.cs
@@ -379,12 +379,12 @@ namespace Xtensive.Sql.Drivers.SqlServer.v09
 
     #region Static helpers
 
-    protected static SqlCast CastToLong(SqlExpression arg)
+    private static SqlCast CastToLong(SqlExpression arg)
     {
       return SqlDml.Cast(arg, SqlType.Int64);
     }
 
-    protected static SqlCast CastToDecimal(SqlExpression arg, short precision, short scale)
+    private static SqlCast CastToDecimal(SqlExpression arg, short precision, short scale)
     {
       return SqlDml.Cast(arg, SqlType.Decimal, precision, scale);
     }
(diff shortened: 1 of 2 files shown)


Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v09/Compiler.cs
```diff
@@ -379,12 +379,12 @@ public override void Visit(SqlCreateIndex node, IndexColumn item)
 
     #region Static helpers
 
-    private static SqlCast CastToLong(SqlExpression arg)
+    protected static SqlCast CastToLong(SqlExpression arg)
```
AlexUstinov, line 382: Why the access modifier has changed to `protected` here?

Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v09/Compiler.cs
```diff
@@ -379,12 +379,12 @@ public override void Visit(SqlCreateIndex node, IndexColumn item)
...
     {
       return SqlDml.Cast(arg, SqlType.Int64);
     }
 
-    private static SqlCast CastToDecimal(SqlExpression arg, short precision, short scale)
+    protected static SqlCast CastToDecimal(SqlExpression arg, short precision, short scale)
```
AlexUstinov, line 387: Why the access modifier has changed to `protected` here?

Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v09/Compiler.cs
alex-kulakov: This and the method bellow left protected accidentally after experimenting with original formula. There is nothing criminal since these methods are accessible within SqlServer providers but I revert them back to private

PR #207, opened 2021-12-10
Minor optimizations (#59)

SergeiPavlov:
* Make QueryTranslationResult readonly struct

* Make QueryRequest readonly struct

* Don't load string resource until necessary

* improve is/as usages

* Optimize PrefetchHelper && make Fetcher readonluy struct

* Optimize BatchingCommandProcessor.PutTasksForExecution

* Refactor PrefetchManager.GetGraphContainer() to avoid unnecessary allocation

* Avoid double HashSet-search in TypeRegistry.Register()

* Improving refactoring of TypeDefCollection

* Use Queue.TryDequeue()

* Avoid ReadOnly-wrapping of Empty arrays. They are always immutable

* Avoid double dictionary search when removing

* Make BuildBatch() to accept IReadOnlyList<string> arg to avoid conversions

* Make ColumnInfoRef readonly struct

* Refactor ColumnCollection

* Make TypeReference readonly struct

* Make ColumnGroup readonly struct

* Make TypeInfoRef readonly struct

### DIFF
--- a/Orm/Xtensive.Orm/Orm/Domain.cs
+++ b/Orm/Xtensive.Orm/Orm/Domain.cs
@@ -119,8 +119,9 @@ namespace Xtensive.Orm
 
     internal KeyGeneratorRegistry KeyGenerators { get; private set; }
 
-    internal ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>> PrefetchFieldDescriptorCache { get; private set; }
-    
+    internal ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>> PrefetchFieldDescriptorCache { get; } =
+       new ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>>();
+
     internal ICache<object, Pair<object, ParameterizedQuery>> QueryCache { get; private set; }
 
     internal ICache<Key, Key> KeyCache { get; private set; }
@@ -420,7 +421,6 @@ namespace Xtensive.Orm
       GenericKeyFactories = new ConcurrentDictionary<TypeInfo, GenericKeyFactory>();
       EntityDataReader = new EntityDataReader(this);
       KeyGenerators = new KeyGeneratorRegistry();
-      PrefetchFieldDescriptorCache = new ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>>();
       KeyCache = new LruCache<Key, Key>(Configuration.KeyCacheSize, k => k);
       QueryCache = new FastConcurrentLruCache<object, Pair<object, ParameterizedQuery>>(Configuration.QueryCacheSize, k => k.First);
       PrefetchActionMap = new Dictionary<TypeInfo, Action<SessionHandler, IEnumerable<Key>>>();
--- a/Orm/Xtensive.Orm/Orm/Domain.cs
+++ b/Orm/Xtensive.Orm/Orm/Domain.cs
@@ -119,8 +119,7 @@ namespace Xtensive.Orm
 
     internal KeyGeneratorRegistry KeyGenerators { get; private set; }
 
-    internal ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>> PrefetchFieldDescriptorCache { get; } =
-       new ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>>();
+    internal ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>> PrefetchFieldDescriptorCache { get; }
 
     internal ICache<object, Pair<object, ParameterizedQuery>> QueryCache { get; private set; }
 
@@ -421,6 +420,7 @@ namespace Xtensive.Orm
       GenericKeyFactories = new ConcurrentDictionary<TypeInfo, GenericKeyFactory>();
       EntityDataReader = new EntityDataReader(this);
       KeyGenerators = new KeyGeneratorRegistry();
+      PrefetchFieldDescriptorCache = new ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>>();
       KeyCache = new LruCache<Key, Key>(Configuration.KeyCacheSize, k => k);
       QueryCache = new FastConcurrentLruCache<object, Pair<object, ParameterizedQuery>>(Configuration.QueryCacheSize, k => k.First);
       PrefetchActionMap = new Dictionary<TypeInfo, Action<SessionHandler, IEnumerable<Key>>>();
--- a/Orm/Xtensive.Orm/Orm/Linq/Translator.Expressions.cs
+++ b/Orm/Xtensive.Orm/Orm/Linq/Translator.Expressions.cs
@@ -702,12 +702,12 @@ namespace Xtensive.Orm.Linq
       var leftIsConstant = context.Evaluator.CanBeEvaluated(left);
       bool leftOrRightIsIndex = false;
 
-      if (left is IndexExpression) {
-        left = VisitIndex((IndexExpression) left);
+      if (left is IndexExpression leftIndexExpression) {
+        left = VisitIndex(leftIndexExpression);
         leftOrRightIsIndex = true;
       }
-      if (right is IndexExpression) {
-        right = VisitIndex((IndexExpression) right);
+      if (right is IndexExpression rightIndexExpression) {
+        right = VisitIndex(rightIndexExpression);
         leftOrRightIsIndex = true;
       }
 
@@ -975,12 +975,13 @@ namespace Xtensive.Orm.Linq
       Type structureType)
     {
       expression = expression.StripCasts();
-      if (expression is IPersistentExpression)
-        return ((IPersistentExpression) expression)
+      if (expression is IPersistentExpression persistentExpression) {
+        return persistentExpression
           .Fields
           .Where(field => field.GetMemberType()==MemberType.Primitive)
           .Select(e => (Expression) e)
           .ToList();
+      }
 
       ConstantExpression nullExpression = Expression.Constant(null, structureType);
       BinaryExpression isNullExpression = Expression.Equal(expression, nullExpression);
@@ -1036,8 +1037,9 @@ namespace Xtensive.Orm.Linq
     private static IList<Expression> GetEntityFields(Expression expression, IEnumerable<Type> keyFieldTypes)
     {
       expression = expression.StripCasts();
-      if (expression is IEntityExpression)
-        return GetKeyFields(((IEntityExpression) expression).Key, null);
+      if (expression is IEntityExpression entityExpression) {
+        return GetKeyFields(entityExpression.Key, null);
+      }
 
 
       Expression keyExpression;
--- a/Orm/Xtensive.Orm/Orm/Linq/Translator.Expressions.cs
+++ b/Orm/Xtensive.Orm/Orm/Linq/Translator.Expressions.cs
@@ -1594,7 +1594,10 @@ namespace Xtensive.Orm.Linq
     {
       var storedEntityType = translator.State.TypeOfEntityStoredInKey;
       var itemToTupleConverter = ItemToTupleConverter.BuildConverter(itemType, storedEntityType, value, translator.context.Model, sourceExpression);
-      var rsHeader = new RecordSetHeader(itemToTupleConverter.TupleDescriptor, itemToTupleConverter.TupleDescriptor.Select(x => new SystemColumn(translator.context.GetNextColumnAlias(), 0, x)).Cast<Column>());
+      var rsHeader = new RecordSetHeader(
+        itemToTupleConverter.TupleDescriptor,
+        itemToTupleConverter.TupleDescriptor.Select(x => new SystemColumn(translator.context.GetNextColumnAlias(), 0, x)).Cast<Column>().ToArray(itemToTupleConverter.TupleDescriptor.Count)
+      );
       var rawProvider = new RawProvider(rsHeader, itemToTupleConverter.GetEnumerable());
       var recordset = new StoreProvider(rawProvider);
       var itemProjector = new ItemProjectorExpression(itemToTupleConverter.Expression, recordset, translator.context);
--- a/Orm/Xtensive.Orm/Orm/Linq/Translator.Expressions.cs
+++ b/Orm/Xtensive.Orm/Orm/Linq/Translator.Expressions.cs
@@ -1594,10 +1594,12 @@ namespace Xtensive.Orm.Linq
     {
       var storedEntityType = translator.State.TypeOfEntityStoredInKey;
       var itemToTupleConverter = ItemToTupleConverter.BuildConverter(itemType, storedEntityType, value, translator.context.Model, sourceExpression);
-      var rsHeader = new RecordSetHeader(
-        itemToTupleConverter.TupleDescriptor,
-        itemToTupleConverter.TupleDescriptor.Select(x => new SystemColumn(translator.context.GetNextColumnAlias(), 0, x)).Cast<Column>().ToArray(itemToTupleConverter.TupleDescriptor.Count)
-      );
+      var tupleDescriptor = itemToTupleConverter.TupleDescriptor;
+      var columns = tupleDescriptor
+        .Select(x => new SystemColumn(translator.context.GetNextColumnAlias(), 0, x))
+        .Cast<Column>()
+        .ToArray(tupleDescriptor.Count);
+      var rsHeader = new RecordSetHeader(tupleDescriptor, columns);
       var rawProvider = new RawProvider(rsHeader, itemToTupleConverter.GetEnumerable());
       var recordset = new StoreProvider(rawProvider);
       var itemProjector = new ItemProjectorExpression(itemToTupleConverter.Expression, recordset, translator.context);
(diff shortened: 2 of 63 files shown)


alex-kulakov:
A lot of tests failed. For example, in ```Xtensive.Orm.Tests.Linq```  namespace there are new 172 failed tests, a lot of tests from other namespaces start failing too.

SergeiPavlov:
> A lot of tests failed. For example, in `Xtensive.Orm.Tests.Linq` namespace there are new 170 failed tests, a lot of tests from other namespaces start failing too.

Fixed the bug

Orm/Xtensive.Orm/Orm/Rse/Providers/Compilable/OrderProviderBase.cs
```diff
@@ -66,11 +66,10 @@ protected override void Initialize()
       var comparisonRules = new ComparisonRules[Order.Count];
       for (int i = 0; i < Order.Count; i++) {
         var orderItem = Order[i];
-        var culture = CultureInfo.InvariantCulture;
         var column = Header.Columns[orderItem.Key];
-        var mappedColumn = column as MappedColumn;
-        if (mappedColumn != null && mappedColumn.ColumnInfoRef != null)
-          culture = mappedColumn.ColumnInfoRef.CultureInfo;
+        var culture = column is MappedColumn mappedColumn && mappedColumn.ColumnInfoRef != default
```
AlexUstinov, line 70: Am I right this creates memory traffic by making copies of ColumnInfoRef to send it to the inequality operator and then to the Equals method?

SergeiPavlov: Yes, you are right
optimized it

Orm/Xtensive.Orm/Orm/Rse/Providers/Compilable/OrderProviderBase.cs
```diff
@@ -66,11 +66,11 @@ protected override void Initialize()
       var comparisonRules = new ComparisonRules[Order.Count];
       for (int i = 0; i < Order.Count; i++) {
         var orderItem = Order[i];
-        var culture = CultureInfo.InvariantCulture;
         var column = Header.Columns[orderItem.Key];
-        var mappedColumn = column as MappedColumn;
-        if (mappedColumn != null && mappedColumn.ColumnInfoRef != null)
-          culture = mappedColumn.ColumnInfoRef.CultureInfo;
+        var columnInfoRef = (column as MappedColumn)?.ColumnInfoRef;
```
AlexUstinov, line 70: Can we make `ColumnInfoRef` a `ref readonly` property? So the code would be
```csharp
// Declaration
public class MappedColumn : Column
{
  private readonly ColumnInfoRef _columnInfo;
  public ref readonly ColumnInfoRef ColumnInfoRef => ref _columnInfo;
}

// Usage
var column = Header.Colulmns[orderItem.Key];
var culture = (column is MappedColumn mappedColumn ? mappedColumn.ColumnInfoRef.Culture : null) ?? CultureInfo.InvariantCulture;
```

Orm/Xtensive.Orm/Orm/Model/ColumnInfoRef.cs
```diff
@@ -18,52 +18,40 @@ namespace Xtensive.Orm.Model
...
-    public string ColumnName { get; private set; }
+    public string ColumnName { get; }
 
     /// <summary>
     /// Gets <see cref="CultureInfo"/> info of the <see cref="ColumnInfo"/>.
     /// </summary>
-    public CultureInfo CultureInfo { get; private set; }
+    public CultureInfo CultureInfo { get; }
+
+    public bool HasValue => TypeName != null;
```
AlexUstinov, line 43: There is no XML comment here.
And as shown in other comments we most likely don't need this property.

Orm/Xtensive.Orm/Orm/Rse/Providers/Compilable/OrderProviderBase.cs
```diff
@@ -66,11 +66,11 @@ protected override void Initialize()
...
+        var culture = columnInfoRef?.HasValue == true
+            ? columnInfoRef.Value.CultureInfo
+            : CultureInfo.InvariantCulture;
```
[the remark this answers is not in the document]
SergeiPavlov: I thought about this case, but it was the previous behavior, and I tried to keep it.

Orm/Xtensive.Orm/Orm/Rse/Providers/Compilable/OrderProviderBase.cs
SergeiPavlov: done

Orm/Xtensive.Orm/Orm/Model/ColumnInfoRef.cs
SergeiPavlov: removed it

alex-kulakov:
Tests are still failing. Take a look at the PrefetchManagerBasicTests with InvalidCastException
```
Unable to cast object of type System.Collections.Generic.Dictionary<System.ValueTuple<Xtensive.Orm.Key,Xtensive.Orm.Model.TypeInfo>,Xtensive.Orm.Internals.Prefetch.GraphContainer> to type <System.Collections.Generic.IEnumerable<Xtensive.Orm.Internals.Prefetch.GraphContainer>.
```

The PrefetchManagerAdvancedTest  - the same exception (or similar, I just glanced at them, definitely InvalidCastException).

SergeiPavlov:
Fixed `InvalidCastException`.  All of them because dynamic casting of `GraphContainersField`, which implementation type changed

Orm/Xtensive.Orm/Orm/Domain.cs
```diff
@@ -119,8 +119,9 @@ public static Domain Demand()
 
     internal KeyGeneratorRegistry KeyGenerators { get; private set; }
 
-    internal ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>> PrefetchFieldDescriptorCache { get; private set; }
-    
+    internal ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>> PrefetchFieldDescriptorCache { get; } =
+       new ConcurrentDictionary<TypeInfo, IReadOnlyList<PrefetchFieldDescriptor>>();
+
```
alex-kulakov, line 124: Why don't you just remove ```private set``` part?

alex-kulakov (changes requested):

Orm/Xtensive.Orm/Orm/Domain.cs
SergeiPavlov: in-place initialization improves readability

Orm/Xtensive.Orm/Orm/TypeReference.cs
```diff
@@ -14,17 +14,17 @@ namespace Xtensive.Orm
...
     /// Gets or sets the referenced type.
     /// </summary>
-    public TypeInfo Type { get; private set; }
+    public TypeInfo Type { get; }
 
     /// <summary>
     /// Gets or sets the type reference accuracy.
     /// </summary>
-    public TypeReferenceAccuracy Accuracy { get; private set; }
+    public TypeReferenceAccuracy Accuracy { get; }
```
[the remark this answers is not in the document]
SergeiPavlov: done

Orm/Xtensive.Orm/Orm/TypeReference.cs
```diff
@@ -14,17 +14,17 @@ namespace Xtensive.Orm
...
   /// </summary>
   [Serializable]
-  public struct TypeReference
+  public readonly struct TypeReference
   {
     /// <summary>
     /// Gets or sets the referenced type.
     /// </summary>
-    public TypeInfo Type { get; private set; }
+    public TypeInfo Type { get; }
```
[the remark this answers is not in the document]
SergeiPavlov: done

Orm/Xtensive.Orm/Orm/Rse/MappedColumn.cs
```diff
@@ -17,10 +17,11 @@ public sealed class MappedColumn : Column
   {
     private const string ToStringFormat = "{0} = {1}";
 
+    private readonly ColumnInfoRef _columnInfo;
```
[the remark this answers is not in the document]
SergeiPavlov: fixed

Orm/Xtensive.Orm/Orm/Services/QueryBuilding/QueryTranslationResult.cs
```diff
@@ -1,36 +1,37 @@
...
     /// Gets or sets SQL DOM query.
     /// </summary>
-    public ISqlCompileUnit Query { get; set; }
+    public SqlSelect Query { get; }
 
     /// <summary>
     /// Gets or sets parameter bindings for translated query.
     /// </summary>
-    public IList<QueryParameterBinding> ParameterBindings { get; set; }
+    public IReadOnlyList<QueryParameterBinding> ParameterBindings { get; }
```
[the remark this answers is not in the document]
SergeiPavlov: done

Orm/Xtensive.Orm/Orm/Services/QueryBuilding/QueryTranslationResult.cs
```diff
@@ -1,36 +1,37 @@
...
   /// Result of LINQ query translation.
   /// </summary>
-  public sealed class QueryTranslationResult
+  public readonly struct QueryTranslationResult
   {
     /// <summary>
     /// Gets or sets SQL DOM query.
     /// </summary>
-    public ISqlCompileUnit Query { get; set; }
+    public SqlSelect Query { get; }
```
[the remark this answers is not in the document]
SergeiPavlov: done

Orm/Xtensive.Orm/Orm/Rse/ColumnCollection.cs
```diff
@@ -72,36 +66,24 @@ public ColumnCollection Alias(string alias)
     /// <summary>
     /// Returns an enumerator that iterates through the <see href="ColumnCollection"/>.
     /// </summary>
-    public List<Column>.Enumerator GetEnumerator() => columns.GetEnumerator();
```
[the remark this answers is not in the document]
SergeiPavlov: I didn't know about previous optimization.
The new approach looks more consistent. E.g. in the following code:
https://github.com/DataObjects-NET/dataobjects-net/blob/master/Orm/Xtensive.Orm/Orm/Rse/RecordSetHeader.cs#L85

We give `ColumnCollection` which implements `IReadOnlyList` as argument to  ColumnCollection constructor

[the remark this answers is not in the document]
SergeiPavlov: Added inner struct `ColumnCollection.Enumeratory` (similar to `List<>.Enumerator`)

Orm/Xtensive.Orm/Orm/Domain.cs
alex-kulakov: In general yes but what does this another mentioning of property type contributes to readability?

SergeiPavlov: fewer lines of code.
Are you suggesting to put it back into constructor?

Orm/Xtensive.Orm/Orm/Rse/ColumnCollection.cs
```diff
@@ -72,36 +106,29 @@ public ColumnCollection Alias(string alias)
...
     /// <summary>
     /// Initializes a new instance of this class.
     /// </summary>
     /// <param name="columns">Collection of items to add.</param>
-    public ColumnCollection(List<Column> columns)
+    public ColumnCollection(IEnumerable<Column> columns)
     {
-      this.columns = columns;
-      var count = columns.Count;
+      this.columns = columns as IReadOnlyList<Column> ?? columns.ToList();
```
alex-kulakov, line 127: What about not having ```isinst``` check? RecordSetHeader can have ```IReadOnlyList``` instead of ```IEnumerable``` for ```columns``` parameter, right? I counted 3 places when we need to add ```.ToArray(count)```, enumeration is done anyway but we could make it more efficient by using ```.ToArray(count)``` instead of ```.ToList()``` here.

We still would have to have two ```ctors``` but having ready-to-go parameter is better than having computation. every time. Also as I see the only places where columns has to be given as IEnumerable are within ColumnCollection class.

Make sense?

Orm/Xtensive.Orm/Orm/Domain.cs
SergeiPavlov: reverted it

Orm/Xtensive.Orm/Orm/Rse/ColumnCollection.cs
SergeiPavlov: I agree.  Done

Orm/Xtensive.Orm/Orm/Linq/Translator.Expressions.cs
```diff
@@ -1594,7 +1594,10 @@ private static ProjectionExpression CreateLocalCollectionProjectionExpression(Ty
     {
       var storedEntityType = translator.State.TypeOfEntityStoredInKey;
       var itemToTupleConverter = ItemToTupleConverter.BuildConverter(itemType, storedEntityType, value, translator.context.Model, sourceExpression);
-      var rsHeader = new RecordSetHeader(itemToTupleConverter.TupleDescriptor, itemToTupleConverter.TupleDescriptor.Select(x => new SystemColumn(translator.context.GetNextColumnAlias(), 0, x)).Cast<Column>());
+      var rsHeader = new RecordSetHeader(
+        itemToTupleConverter.TupleDescriptor,
+        itemToTupleConverter.TupleDescriptor.Select(x => new SystemColumn(translator.context.GetNextColumnAlias(), 0, x)).Cast<Column>().ToArray(itemToTupleConverter.TupleDescriptor.Count)
+      );
```
alex-kulakov, line 1600: I really think we should use variables like so
```csharp
var tupleDescriptor = itemToTupleConverter.TupleDescriptor;
var columns = tupleDescriptor
  .Select(x => new SystemColumn(translator.context.GetNextColumnAlias(), 0, x))
  .Cast<Column>()
  .ToArray(tupleDescriptor.Count);
var rsHeader = new RecordSetHeader(tupleDescriptor, columns);
```

SergeiPavlov: I agree. done

Orm/Xtensive.Orm/Orm/Rse/Providers/Compilable/OrderProviderBase.cs
```diff
@@ -66,11 +66,8 @@ protected override void Initialize()
       var comparisonRules = new ComparisonRules[Order.Count];
       for (int i = 0; i < Order.Count; i++) {
         var orderItem = Order[i];
-        var culture = CultureInfo.InvariantCulture;
         var column = Header.Columns[orderItem.Key];
-        var mappedColumn = column as MappedColumn;
-        if (mappedColumn != null && mappedColumn.ColumnInfoRef != null)
-          culture = mappedColumn.ColumnInfoRef.CultureInfo;
+        var culture = (column as MappedColumn)?.ColumnInfoRef.CultureInfo ?? CultureInfo.InvariantCulture;
```
alex-kulakov, line 70: This might be a buggy line. You've changed the algorithm and ```null```s are not allowed to reach ```ComparisonRule```.  Some built-in comparison rules have ```CultureInfo``` set to ```null``` so ```null``` culture is valid value, I see it is taken to account all over the ```ComparisonRule``` class. ColumnInfoRef.CultureInfo can be null when ColumnInfo.CultureInfo is null which is theoretically possible but doesn't happen now, I'm not sure it should be but still. Did you see this?

If this causes problem it will be hard to find after few years.

SergeiPavlov: Yes, I thought about this semantic change.
Because `ColumnInfoRef` is converted to struct we cannot keep previous null-check.

But I tracked how `ColumnInfo.CultureInfo` may be set to `null`.  The only ways is setter property, which never called

alex-kulakov: > The only way is property setter, which never called.

This is why I call it "hard to find". :) If somebody who doesn't know this thing or just not pays attention just set the property it will silently create a problem.

> Also we can change property type to nullable ```ColumnInfoRef? ColumnInfoRef.```
Is Checking whether ColumnInfoRef is default instance worse? Like so:
```
var culture = column is MappedColumn mColumn && mColumn.ColumnInfoRef != default
  ? mColumn.ColumnInfoRef.CultureInfo
  : CultureInfo.InvariantCulture;
```

I highly doubt that names of ColumnInfoRef that are both null will be valid case, right?

SergeiPavlov: `mColumn.ColumnInfoRef != default` is expension operation. It invokes `operator !=`, `ColumnInfoRef.Equals`.  

But we can check `ColumnInfoRef .TypeName != null`

SergeiPavlov: > var culture = column is MappedColumn mColumn && mColumn.ColumnInfoRef != default
>   ? mColumn.ColumnInfoRef.CultureInfo
>   : CultureInfo.InvariantCulture;

This was the first solution, but @AlexUstinov  argued against it https://github.com/DataObjects-NET/dataobjects-net/pull/207#discussion_r773553326

alex-kulakov (approved):

Issue #205, opened 2021-12-09
FullText - Azure SQL Server - Issue on building domain, extracted schema differs from targeted schema


rlesavre:
When using a Fulltext index on the model using an Azure SQL server on azure, the initial domain creation works, and the full text index is working however, on next domain building, DO seems to be unable to see the Index, and detect a domain change:

![image](https://user-images.githubusercontent.com/5682585/145376990-cc3b2a9a-da2d-4b48-9563-7f20eabded3b.png)
![image](https://user-images.githubusercontent.com/5682585/145377083-89731420-0d90-4f10-a66a-f8e62ff83b32.png)

Schema comparison result: TargetIsSuperset
Has unsafe actions: false
Has column type changes: false 
Compatible in ValidateLegacy mode: unknown

Unsafe actions:

Schema hints:
  
Schema difference:
  "./" != "./" (NodeDifference): Unknown
    +Tables: "./Tables (7)" != "./Tables (7)" (NodeCollectionDifference): 1 change(s)
      "./Tables/Book" != "./Tables/Book" (NodeDifference): Unknown
        +FullTextIndexes: "./Tables/Book/FullTextIndexes (0)" != "./Tables/Book/FullTextIndexes (1)" (NodeCollectionDifference): 1 change(s)
          "" != "./Tables/Book/FullTextIndexes/FT_Book" (NodeDifference): Created
            +Columns: "" != "./Tables/Book/FullTextIndexes/FT_Book/Columns (1)" (NodeCollectionDifference): 1 change(s)
              "" != "./Tables/Book/FullTextIndexes/FT_Book/Columns/0" (NodeDifference): Created
                +Configuration: "" != "French" (ValueDifference): values differ
                +Value: "" != "./Tables/Book/Columns/Title" (ValueDifference): values differ

This exact same code works on an hosted SQL Server instance, is there is an issue with the extracted Schema ?

alex-kulakov:
Hello @rlesavre,

I don't know such issues yet. Also the DO version is important here, could you specify it? I'll check the case.

rlesavre:
Thanks @alex-kulakov i'm having the same behavior with the 7.0.1 or 6.0.6.

alex-kulakov:
I confirm this problem. I'll fix it in next releases.

Issue #203, opened 2021-12-07
Schema extraction fails when a 32-bit sequence is found


ondrejtucny:
I have a legacy DB, using the LegacySkip upgrade mode. When the Extractor performs ExtractSequences(), it fails with the following exception:

```
Exception “System.AggregateException” with message “One or more errors occurred. (Unable to cast object of type 'System.Int32' to type 'System.Int64'.)” occured; handling ID “9cea9e4a-7c55-4f12-8ba8-6320f237ca23”
<Exception>
    <ExceptionType>System.AggregateException, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e</ExceptionType>
    <Message>One or more errors occurred. (Unable to cast object of type 'System.Int32' to type 'System.Int64'.)</Message>
    <Source>System.Private.CoreLib</Source>
    <HelpLink />
    <Data>
        <HandlingInstanceId>9cea9e4a-7c55-4f12-8ba8-6320f237ca23</HandlingInstanceId>
    </Data>
    <AdditionalInfo>
        <MachineName>CSOB-UC01</MachineName>
        <TimeStamp>2021-07-15 01:25:05.340+02:00</TimeStamp>
        <FullName>Bee.Core, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null</FullName>
        <AppDomainName>DieboldNixdorf.UniCard.Server.Console</AppDomainName>
        <ThreadIdentity>unknown</ThreadIdentity>
    </AdditionalInfo>
    <StackTrace>   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at Xtensive.Core.AsyncFutureResult`1.Get()
   at Xtensive.Orm.Upgrade.UpgradingDomainBuilder.CompleteSqlWorker()
   at Xtensive.Orm.Upgrade.UpgradingDomainBuilder.BuildSingleStageDomain()
   at Xtensive.Orm.Upgrade.UpgradingDomainBuilder.Run()
   at Xtensive.Orm.Upgrade.UpgradingDomainBuilder.Build(DomainConfiguration configuration)
   at Xtensive.Orm.Domain.Build(DomainConfiguration configuration)
[...]

PR #209, opened 2021-12-20
Schema extraction improvements for MS SQL Server

alex-kulakov:
Addresses #203 and #205

- Sequence extraction can extract other types of sequences from ```tinyint``` to ```bigint``` (decimal and numeric are not supported by DO sequence descriptor).
- Full-text indexes are correctly extracted by Azure SQL provider.

### DIFF
--- a/Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v11/Extractor.cs
+++ b/Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v11/Extractor.cs
@@ -50,8 +50,9 @@ namespace Xtensive.Sql.Drivers.SqlServer.v11
           var sequence = currentSchema.CreateSequence(reader.GetString(1));
           var descriptor = sequence.SequenceDescriptor;
 
-          if (!valueReaders.TryGetValue(reader.GetInt32(2), out var valueReader))
-            throw new ArgumentOutOfRangeException($"Type of sequence '{reader.GetString(1)}' is not supported");
+          if (!valueReaders.TryGetValue(reader.GetInt32(2), out var valueReader)) {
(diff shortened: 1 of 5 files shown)


Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v11/Extractor.cs
```diff
@@ -46,12 +49,17 @@ ORDER BY
...
           var descriptor = sequence.SequenceDescriptor;
-          descriptor.StartValue = reader.GetInt64(2);
-          descriptor.Increment = reader.GetInt64(3);
-          descriptor.MinValue = reader.GetInt64(4);
-          descriptor.MaxValue = reader.GetInt64(5);
-          descriptor.IsCyclic = reader.GetBoolean(6);
-          descriptor.LastValue = reader.GetInt64(7);
+
+          if (!valueReaders.TryGetValue(reader.GetInt32(2), out var valueReader)) {
+            throw new ArgumentOutOfRangeException($"Type of sequence '{reader.GetString(1)}' is not supported.");
```
AlexUstinov, line 54: We may have `reader.GetString(1)` as a variable. Now code assumes the current query result row is cached in memory, so it reads column with index 1 again after the column with the index 2 is already read. However, in general, that assumption is not always true. For example `cmd.ExecuteReader(CommandBehavior.SequentialAccess)` in theory must be faster but it doesn't allow stepping back.

AlexUstinov (approved):

Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v11/Extractor.cs
alex-kulakov: Sounds reasonable, so I'll make a variable for it.

PR #210, opened 2022-01-03
Disable BooleanExpressionConverter optimization for .OrderBy() and .ThenBy() expressions

SergeiPavlov:
The optimization works incorrectly for expressions like .OrderBy(o => o.Status == 1)

The fix disables the optimization for Sorting expressions
Added OrderByBoolExpression test to check this case

### DIFF
--- a/Orm/Xtensive.Orm.Tests/Linq/OrderByTest.cs
+++ b/Orm/Xtensive.Orm.Tests/Linq/OrderByTest.cs
@@ -316,5 +316,14 @@ namespace Xtensive.Orm.Tests.Linq
       Assert.That(result, Is.Not.Empty);
       Assert.IsTrue(expected.SequenceEqual(result));
     }
+
+    [Test]
+    public void OrderByBoolExpression()
+    {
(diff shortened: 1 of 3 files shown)


alex-kulakov:
Hello @SergeiPavlov 

First of all, the problem originally appeared in ```7.0``` so fixing it in master now is not an option. Also not only ```order by``` is affected. My idea is similar to yours but since the root of the problem is ```enums``` I try to turn off ```CASE``` omit only if enum detected in calculated columns, it is also not so flexible thought. I continue investigation.

alex-kulakov:
Wait for some time, I'm testing the changes from the Denis' PR now on different storages

alex-kulakov:
@SergeiPavlov, the servicetitan#61 seems to work fine on several storages, I will run tests on the all storage configs for actual PR for 7.0

SergeiPavlov:
> @SergeiPavlov, the [servicetitan#61](https://github.com/servicetitan/dataobjects-net/pull/61) seems to work fine on several storages, I will run tests on the all storage configs for actual PR for 7.0

Replaced by https://github.com/DataObjects-NET/dataobjects-net/pull/212

Issue #61, opened 2020-09-10
FieldInfo.IsDynamicallyDefiled field has mistyping in it


alex-kulakov:
Now its name is IsDynalicallyDefined which is not OK.

PR #212, opened 2022-01-11
BooleanExpressionConverter: use precise optimization (#61)

SergeiPavlov:
* BooleanExpressionConverter: use precise optimization
* BaseSqlVisitor: handle SqlMetadata
* add more tests for handling boolean expressions

### DIFF
--- a/Orm/Xtensive.Orm/Sql/Dml/Expressions/SqlMetadata.cs
+++ b/Orm/Xtensive.Orm/Sql/Dml/Expressions/SqlMetadata.cs
@@ -1,0 +1,39 @@
+using Xtensive.Core;
+
+namespace Xtensive.Sql.Dml
+{
+  /// <summary>
+  /// Arbitrary metadata that could be attached to SQL expression tree.
+  /// </summary>
+  public class SqlMetadata : SqlExpression
+  {
+    public SqlExpression Expression { get; private set; }
+
+    public object Value { get; private set; }
+
+    public override void ReplaceWith(SqlExpression expression)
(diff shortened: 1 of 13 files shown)


alex-kulakov:
Seems to be that due to some differences in ```order by``` work across RDBMSs the ```OrderByTest.OrderByBoolExpressionComplex``` test fails in Firebird, MySQL 5.5 and all supported PostreSql versions. Translated query is the same so it is definitely RDBMS behavior but I would have no new failing tests.

The check of last item in the array fails, MS SQL Server returns results that end with ```Paid``` items followed by  ```Unpaid``` item, and the storages above return results that end with ```Unpaid``` items followed by ```Paid``` items (they are switched). I think test should be corrected to work across all storages.

SergeiPavlov:
Fixed the undefined behavior in `OrderByTest.OrderByBoolExpressionComplex`

Orm/Xtensive.Orm/Sql/Dml/Expressions/SqlMetadata.cs
```diff
@@ -0,0 +1,39 @@
+using Xtensive.Core;
```
alex-kulakov, line 1:
```suggestion
// Copyright (C) 2022 Xtensive LLC.
// This code is distributed under MIT license terms.
// See the License.txt file in the project root for more information.

using Xtensive.Core;
```

alex-kulakov (approved):

alex-kulakov (approved):

PR #214, opened 2022-01-21
Sql Server v13+ Datetime subtraction issue fix

alex-kulakov:
Previous movement to ```DATEDIFF_BIG``` (#198) appeared to have problems with some ```DateTime``` values.

This PR tries to keep better performance of less calculations while not sacrificing full range of dates to be supported. It brings back previous formula of date diff calculation for some scenarios (```DATEDIFF_BIG``` can't handle diff in nanoseconds fully, [more info here](https://docs.microsoft.com/ru-ru/sql/t-sql/functions/datediff-big-transact-sql?view=sql-server-ver15#return-value)), while using relatively lightweight date diff calculation for the most part.

### DIFF
--- a/Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v09/Compiler.cs
+++ b/Orm/Xtensive.Orm.SqlServer/Sql.Drivers.SqlServer/v09/Compiler.cs
@@ -18,6 +18,18 @@ namespace Xtensive.Sql.Drivers.SqlServer.v09
 {
   internal class Compiler : SqlCompiler
   {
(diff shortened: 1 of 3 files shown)


AlexUstinov:
Is there a test for the overflow case? Could you please add one if it doesn't exist?

alex-kulakov:
There is a test ```IssueJira0556_ArithmeticExceptionWhenSubstractionOfDates.cs``` , I added it long ago. I didn't notice it became failed so this PR fixes my mistake in #198.

AlexUstinov (approved):

PR #218, opened 2022-02-03
A bunch of small changes

alex-kulakov:
- ```SimpleCommandProcessor``` manages task queue the same way as ```BatchingCommandProcessor``` (more efficient).
- Correct errors for ```.Tag()``` query extension if base query provider is not ours.
- T4 templates are correctly assigned with the files that they generate.
- Some notes in ColumnCollection and ColumnGroupCollection about unsafe optimization (just in case, to lower chances of wrong usage)
- Got rid of a lot of error messages which weren't in use.

### DIFF
--- a/Orm/Xtensive.Orm/Orm/Providers/CommandProcessing/SimpleCommandProcessor.cs
+++ b/Orm/Xtensive.Orm/Orm/Providers/CommandProcessing/SimpleCommandProcessor.cs
@@ -14,7 +14,7 @@ namespace Xtensive.Orm.Providers
 {
   internal sealed class SimpleCommandProcessor : CommandProcessor, ISqlTaskProcessor
   {
-    private readonly Queue<SqlTask> tasks;
+    private Queue<SqlTask> tasks = new();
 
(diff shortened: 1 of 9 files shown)


Orm/Xtensive.Orm/Orm/Providers/CommandProcessing/SimpleCommandProcessor.cs
```diff
@@ -51,8 +51,8 @@ void ISqlTaskProcessor.ProcessTask(SqlPersistTask task, CommandProcessorContext
 
     public override void ExecuteTasks(CommandProcessorContext context)
     {
-      context.ProcessingTasks = new Queue<SqlTask>(tasks);
-      tasks.Clear();
+      context.ProcessingTasks = tasks;
+      tasks = new Queue<SqlTask>();
```
AlexUstinov, line 55: I'd think about increasing the initial Queue capacity to reduce the reallocation rate.

AlexUstinov (approved):

AlexUstinov (approved):

PR #219, opened 2022-02-28
Throw exception on unsupported IQueryable.DistinctBy() expressions

SergeiPavlov:


Orm/Xtensive.Orm/Orm/Linq/Translator.Queryable.cs
```diff
@@ -132,6 +132,8 @@ protected override Expression VisitQueryableMethod(MethodCallExpression mc, Quer
             }
 
             break;
+          case QueryableMethodKind.DistinctBy:
+            throw new NotSupportedException(Strings.UnsupportedDistinctBy);
```
alex-kulakov, line 136: Please, add 'Ex' prefix for the exception message title in resources, by such prefix we differentiate exception messages from any other messages. So in this case the ```ExUnsupportedDistinctBy``` is preferred name.

alex-kulakov (changes requested):

Orm/Xtensive.Orm/Orm/Linq/Translator.Queryable.cs
SergeiPavlov: done

alex-kulakov (approved):

Issue #224, opened 2022-03-08
QueryCache keeps reference to object in closure


SergeiPavlov:
It is expected that QueryCache should not keep references to any objects from closure.
Because cache record lifetime may be indefinitely long

See test & fix in https://github.com/DataObjects-NET/dataobjects-net/pull/223

PR #223, opened 2022-03-08
Issue: Cached query captures object in closure

SergeiPavlov:
It is expected that QueryCache should not keep references to any objects from closure.
Because cache record lifetime may be indefinitely long

PR #225, opened 2022-03-08
Fix Issue 224 (Caching queries with object captures)

SergeiPavlov:
Fixes https://github.com/DataObjects-NET/dataobjects-net/issues/224


SergeiPavlov:
Replaced by https://github.com/DataObjects-NET/dataobjects-net/pull/223

PR #234, opened 2022-03-31
Consider DO_TargetFrameworks env var to multitargeting build

SergeiPavlov:
For effective usage of .NET6 optimization features (e.g. string interpolation) we need do build DO package .NET6.

With `DO_TargetFrameworks` environment variable we get possibility to tune build as we want:
* By default the solution builds for `net5.0` only as before

* To build for both .NET5 & .NET6 define `DO_TargetFrameworks`
```
SET  DO_TargetFrameworks=net5.0;net6.0
dotnet build   
```

* for development purposes it is recommended to define only single TargetFramework to minimize build time. And define both only for package deployment

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -40,7 +40,8 @@
   <PropertyGroup>
     <NoLogo>true</NoLogo>
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFrameworks>net5.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(DO_TargetFrameworks)' != ''">$(DO_TargetFrameworks)</TargetFrameworks>
     <LangVersion>9.0</LangVersion>
     <SolutionDir Condition="$(SolutionDir) == ''">$([MSBuild]::EnsureTrailingSlash(
       $([MSBuild]::GetDirectoryNameOfFileAbove('$(MSBuildThisFileDirectory)', 'Orm.sln'))))</SolutionDir>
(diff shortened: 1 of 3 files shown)


alex-kulakov:
@SergeiPavlov What if ```TargetFrameworks``` is set as property in ```dotnet build```? Now target framework is set to be net5.0 only, but if we want it to be definable from the outside we should consider property too. Unfortunately, neither git console nor powershell  allows to set  ```-p:TargetFrameworks=net5.0;net6.0``` (I tried but failed, that means it is not intuitive so almost no one will use it even if there is a way) but with one framework it works

What if we had 
1) $(TargetFrameworks) - the property from dotnet build
2) $(DO_TargetFrameworks) - if step 1 ended up with empty string
3) fallback to net5.0 if variable is still empty

In this case smaller scope setting can override global setting, It could be helpful if we need to build projects for only one framework (like on local machine instead of build server).

Makes sense? Would it be beneficial for your case?

SergeiPavlov:
> What if TargetFrameworks is set as property in dotnet build
How it can be set ?

Do you mean following code?:
```
        <TargetFrameworks Condition="'$(TargetFrameworks)'=='' AND '$(DO_TargetFrameworks)'!=''">$(DO_TargetFrameworks)</TargetFrameworks>
        <TargetFrameworks Condition="'$(TargetFrameworks)'=='' AND '$(DO_TargetFrameworks)'==''">net5.0</TargetFrameworks>
```

It seem to be equivalent to this PR.

alex-kulakov:
> How it can be set ?

Like so ```dotnet build -c Release -p:TargetFrameworks=net6.0```

> Do you mean following code?:

This will probably work
```
<TargetFrameworks>$(TargetFrameworks)</TargetFrameworks> <!-- the property -->
<TargetFrameworks Condition="'$(TargetFrameworks)'==''">$(DO_TargetFrameworks)</TargetFrameworks> <!-- env var -->
<TargetFrameworks Condition="'$(TargetFrameworks)'==''">net5.0</TargetFrameworks> <!-- fallback to default -->
```

SergeiPavlov:
Yes, it does work
Updated the PR

alex-kulakov (approved):

alex-kulakov:
Seems to be working. I made comments within TargetFrameworks declaration just for you, it would be just fine without them, not so much people get into this proj/targets/csproj files to do things, when they do, they will probably recognize this property override pattern :)

I will agree to set net6.0 when the we begin to build for it. I this case yes, I'll switch fallback framework to ```net6.0```. For now, default framework should stay ```net5.0```. I do think that we need to add it before next Beta, though, but I need to check that packages are built fine and prepare our test environment for the switch.

PR #236, opened 2022-04-05
Optimization: avoid Lambda JITing for typical case of invoking Bulk Set() with value

SergeiPavlov:
Following code snippet:
```
session.Query.All<User>().Set(u => u.UpdateTime, now).Update();
```
compiles and executes expression lambda `a => now`, while `now` value is already known.
It is significant overhead. We see it in Profiler

### DIFF
--- a/Extensions/Xtensive.Orm.BulkOperations/BulkExtensions.cs
+++ b/Extensions/Xtensive.Orm.BulkOperations/BulkExtensions.cs
@@ -78,11 +78,14 @@ namespace Xtensive.Orm.BulkOperations
     /// <returns>Instance of <see cref=" IUpdatable&lt;T&gt;"/>.</returns>
     [Pure]
     public static IUpdatable<T> Set<T, TResult>(this IQueryable<T> query, Expression<Func<T, TResult>> field,
-        TResult value) where T: IEntity =>
-      Set(query,
-        field,
-        Expression.Lambda<Func<T, TResult>>(Expression.Constant(value, typeof(TResult)), Expression.Parameter(typeof(T), "a"))   // Manually constructed expression is simpler than `a => value`
-      );
+      TResult value) where T: IEntity
+    {
+      // Manually constructed expression is simpler than `a => value`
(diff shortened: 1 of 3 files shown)


botinko (approved):

Extensions/Xtensive.Orm.BulkOperations/BulkExtensions.cs
```diff
@@ -78,8 +78,11 @@ public static Task<int> DeleteAsync<T>(this IQueryable<T> query, CancellationTok
     /// <returns>Instance of <see cref=" IUpdatable&lt;T&gt;"/>.</returns>
     [Pure]
     public static IUpdatable<T> Set<T, TResult>(this IQueryable<T> query, Expression<Func<T, TResult>> field,
-      TResult value) where T: IEntity =>
-      Set(query, field, a => value);
+        TResult value) where T: IEntity =>
+      Set(query,
+        field,
+        Expression.Lambda<Func<T, TResult>>(Expression.Constant(value, typeof(TResult)), Expression.Parameter(typeof(T), "a"))   // Manually constructed expression is simpler than `a => value`
+      );
```
alex-kulakov, line 85:
```suggestion
      TResult value) where T: IEntity
    {
      // Manually constructed expression is simpler than `a => value`
      var valueFunc = Expression.Lambda<Func<T, TResult>>(Expression.Constant(value, typeof(TResult)),
        Expression.Parameter(typeof(T), "a"));

      return Set(query, field, valueFunc);
    }
```

alex-kulakov, line 85:
```suggestion
      TResult value) where T: IEntity
    {
      // Manually constructed expression is simpler than `a => value`
      var valueFunc = Expression.Lambda<Func<T, TResult>>(Expression.Constant(value, typeof(TResult)),
        Expression.Parameter(typeof(T), "a"));

      return Set(query, field, valueFunc);
    }
```

alex-kulakov (approved):

botinko (approved):

PR #237, opened 2022-04-09
Optimize work with temporary tables when translating `.In()` extension

SergeiPavlov:
* Clear temporary tables by `TRUNCATE TABLE` statement for efficiency when RDBMS supports it
* Use multi-record `INSERT INTO` statements to fill temporary table.  `SqlInsert` class extended to support such kind of operation.

Benchmarking shows 2x improvement of performance & memory consumption:
```
session.Query.All<Customer>().Where(u => u.Id.In(ids)).ToArray(); 
where `ids` array contains 257 elements (Test257)
& where `ids` array contains 10 elements (Test10)

TRUNCATE TABLE with multi-record INSERT  (batches by 256/16/1 records), Lazy Sql translation
|    Method |       Mean | Error |        Min |        Max |  Gen 0 | Allocated |
|---------- |-----------:|------:|-----------:|-----------:|-------:|----------:|
|  TestIn10 |   794.5 us |    NA |   794.5 us |   794.5 us |      - |    229 KB |
| TestIn257 | 9,870.9 us |    NA | 9,870.9 us | 9,870.9 us | 4.0000 |    558 KB |

TRUNCATE TABLE with multi-record INSERT  (batches by 10 records)
| Method |     Mean | Error |      Min |      Max | Allocated |
|------- |---------:|------:|---------:|---------:|----------:|
| TestIn | 11.53 ms |    NA | 11.53 ms | 11.53 ms |    488 KB |


TRUNCATE TABLE
| Method |     Mean | Error |      Min |      Max | Allocated |
|------- |---------:|------:|---------:|---------:|----------:|
| TestIn | 26.34 ms |    NA | 26.34 ms | 26.34 ms |      1 MB |

DELETE FROM temp
| Method |     Mean | Error |      Min |      Max | Allocated |
|------- |---------:|------:|---------:|---------:|----------:|
| TestIn | 27.28 ms |    NA | 27.28 ms | 27.28 ms |      1 MB |

DROP TABLE
| Method |     Mean | Error |      Min |      Max | Allocated |
|------- |---------:|------:|---------:|---------:|----------:|
|  TestIn10 |    768.7 us |    NA |    768.7 us |    768.7 us |      - |    225 KB |
| TestIn257 | 36,315.3 us |    NA | 36,315.3 us | 36,315.3 us | 8.0000 |  1,392 KB |

```


botinko:
Common suggestion: `INSERT INTO` approach will not work with MARS enabled and XACT_ABORT disabled due to limitations on SQL server side. We need a way to configure `INSERT INTO` approach.

SergeiPavlov:
> Common suggestion: `INSERT INTO` approach will not work with MARS enabled and XACT_ABORT disabled due to limitations on SQL server side. We need a way to configure `INSERT INTO` approach.

Its look like this issue related to `BULK INSERT` operation, but not to multi-record  `INSERT INTO`

botinko:
> > Common suggestion: `INSERT INTO` approach will not work with MARS enabled and XACT_ABORT disabled due to limitations on SQL server side. We need a way to configure `INSERT INTO` approach.
> 
> Its look like this issue related to `BULK INSERT` operation, but not to multi-record `INSERT INTO`

You're right.

Orm/Xtensive.Orm/Orm/WellKnown.cs
```diff
@@ -95,6 +95,11 @@ public static partial class WellKnown
     /// </summary>
     public const int MaxNumberOfConditions = 256;
 
+    /// <summary>
+    /// Number of records in multi-record INSERT SQL statement
+    /// </summary>
+    public const int RecordsInInsertBatch = 10;
```
botinko, line 101: Is 10 a good constant? Why not 42 😄 ? I think we need to make some benchmarking to choose a good default value or convert this constant a configuration.

SergeiPavlov: We can calulate optimimal value. For eaxmple if we have `.In()` with 259 items then:
`RecordsInInsertBatch = 10` case:    25 batches + 9 single queries = 34 INSERTs
`RecordsInInsertBatch = 20` case:    12 batches + 19 single queries = 31 INSERTs

not a big difference.

For better optimization we should precompiles batches with difgferent size: e.g.  8, 16, 32... and combine them.  But then we will have bigger compilation overhead.

I think it would be micro-optimization

alex-kulakov: Have you seen this or something similar? https://www.red-gate.com/simple-talk/databases/sql-server/performance-sql-server/comparing-multiple-rows-insert-vs-single-row-insert-with-three-data-load-methods/

Thought it is quite old it raises question about sweet-spot number of values.

Also we have ```BatchSize``` setting for session (```SessionConfiguration.BatchSize```) and constant value ```SessionConfiguration.DefaultBatchSize = 25```

Number of parameters also very across RDBMSs, we did batch sizing that takes maximum number of parameters into account, it could be important here too

alex-kulakov:
Would be reasonable to implement multi-record PersistRequests first. Now we're doing this kind of backwards. I'll see-thru persist process and be back with suggestions of how we can silver-bullet both scenarios.

SergeiPavlov:
> Would be reasonable to implement multi-record PersistRequests first. Now we're doing this kind of backwards. I'll see-thru persist process and be back with suggestions of how we can silver-bullet both scenarios.
> 
> For now having StoreRequest + BatchStoreRequest at the same time looks weird.

At the time of request creations the code dones not know number of elements. So it cannot create single multi-record request.

Orm/Xtensive.Orm/Orm/WellKnown.cs
SergeiPavlov: Thank you for the link.

When we don't know exact number of elements, the best estimate for batch size would be SQRT(average # elements).
When our threshold for temporary tables if `256`  the relatively efficient batch size should be `>SQRT(256)==16` 

I'm going to make Threshold value configurable instead of hardcoding and to calculate batchsize as SQRT(Threshold)

SergeiPavlov:
The final algorithm:
3-level batching:
* INSERT Batches by 256 records
* then byt 16 records
* then by 1 record.

alex-kulakov:
"Orm\Xtensive.Orm\Sql\Compiler\SqlTranslator.cs(1262,101): error CS1002: ; expected"

SergeiPavlov:
> "Orm\Xtensive.Orm\Sql\Compiler\SqlTranslator.cs(1262,101): error CS1002: ; expected"

The build is fixed

SergeiPavlov:
> * Xtensive.Orm.Tests.Linq.LocalCollectionTest
> * Xtensive.Orm.Tests.Test.Issues.IssueJira570_InOperationDoesNotCreateTemporaryTable

Fixed

SergeiPavlov:
> is actually `ICollection<Tuple>` so we could know the count and use it instead of guessing, couldn't we?
> 
> In items count range from 256 to 4000, probability of having at least one 256-items record is 1, and NOT HAVING 1-item record is 0.0587 so these two requests may be regular, not Lazy, no?

At the time of  `TemporaryTableDescriptor` creating we dont have access to `tuples`. So we cannot chose what to implement: `IPersistDescriptor` or `IMultiRecordPersistDescriptor`. Lazy-ness just a way to postpend the decision until `.Store()` invocation.

alex-kulakov:
Please, write commit messages related to what is in the commit. "Fix tests" has nothing to do with tests, no test was changed, no problem with actual tests fixed, the truth is that you cut off complex structures from multi-row inserts and what should've been written in the commit message. Such messages are misleading and should not exist because they hide true meaning of the changes inside for the history reader.

SergeiPavlov:
Sure.  Usually I don't attach much importance to PR's sub-commit messages.
When PRs will be squashed during merge, they will not be in the final Title message.

SergeiPavlov:
> would extend IPersistDescriptor with something like IMultiRecordPersistDescriptor with new properties, and then make

Finally I got what you mean:
https://github.com/DataObjects-NET/dataobjects-net/pull/237/commits/ad3fc077897d33efb0312d8c7c491148304ce18d

alex-kulakov:
I've been thinking about how to make ```SqlInsert``` multi-row and at the same time convenient to work with. The idea with, basically, ```Dictionary<SqlColumn, List<SqlExpression>> ``` relies on trust to one who uses it and, to me, such design will backfire us with new issues in future. SqlDml is used by not only us but some customers as well.

Here is a case which breaks SqlInsert.Values implemented here:
```csharp
[Test]
public void Test03()
{
  var createTable = "CREATE TABLE [SqlInsertTestTableTest03]" +
                   "( [Id]             integer not null" +
                   ", [FirstName]      nvarchar(255) not null" +
                   ", [LastName]       nvarchar(255) not null" +
                   ", [Birthday]       datetime2 not null)";
  _ = ExecuteNonQuery(createTable);

  var schema = ExtractDefaultSchema();
  var table = schema.Tables["SqlInsertTestTableTest03"];
  var tableRef = SqlDml.TableRef(table);

  var insert = SqlDml.Insert(tableRef);

  insert.Values.Add(tableRef["Id"], SqlDml.Literal(1));
  insert.Values.Add(tableRef["FirstName"], SqlDml.Literal("Alex"));
  insert.Values.Add(tableRef["LastName"], SqlDml.Literal("Kulakov"));
  insert.Values.Add(tableRef["Birthday"], SqlDml.Literal(new DateTime(2008, 10, 12)));

  // inconsistent number of values
  insert.Values.Add(tableRef["Id"], SqlDml.Literal(2));
  //insert.Values.Add(tableRef["FirstName"], SqlDml.Literal("NotAlex"));
  insert.Values.Add(tableRef["LastName"], SqlDml.Literal("NotKulakov"));
  //insert.Values.Add(tableRef["Birthday"], SqlDml.Literal(new DateTime(2008, 10, 14)));

  var compiled = Driver.Compile(insert); // throws ArgumentOutOfRangeException
  var insertCommand = compiled.GetCommandText();
  Console.WriteLine(insertCommand);
  _ = ExecuteNonQuery(insertCommand);
}
```

I think we should explicitly add rows in some way. Old values represented 1 row and it was Dictionary<SqlColumn, SqlExpression>, logic tells that we should use this structure as row in some collection of rows in SqlInsert, at least to take column values from outside. Like so:

```csharp
[Test]
public void Test06()
{
  var createTable = "CREATE TABLE [SqlInsertTestTableTest06]" +
                   "( [Id]             integer not null" +
                   ", [FirstName]      nvarchar(255) not null" +
                   ", [LastName]       nvarchar(255) not null" +
                   ", [Birthday]       datetime2 not null)";
  _ = ExecuteNonQuery(createTable);

  var schema = ExtractDefaultSchema();
  var table = schema.Tables["SqlInsertTestTableTest06"];
  var tableRef = SqlDml.TableRef(table);

  var insert = SqlDml.Insert(tableRef);

  var row = new Dictionary<SqlColumn, SqlExpression>(4);
  row.Add(tableRef["Id"], SqlDml.Literal(1));
  row.Add(tableRef["FirstName"], SqlDml.Literal("Alex"));
  row.Add(tableRef["LastName"], SqlDml.Literal("Kulakov"));
  row.Add(tableRef["Birthday"], SqlDml.Literal(new DateTime(2008, 10, 12)));
  insert.ValueRows.AddRow(row);

  row = new Dictionary<SqlColumn, SqlExpression>(4);
  row.Add(tableRef["Id"], SqlDml.Literal(2));
  row.Add(tableRef["FirstName"], SqlDml.Literal("NotAlex"));
  row.Add(tableRef["LastName"], SqlDml.Literal("NotKulakov"));
  row.Add(tableRef["Birthday"], SqlDml.Literal(new DateTime(2008, 10, 14)));
  insert.ValueRows.AddRow(row);

  var compiled = Driver.Compile(insert);
  var insertCommand = compiled.GetCommandText();
  Console.WriteLine(insertCommand);
  _ = ExecuteNonQuery(insertCommand);
}
```

My raw implementation of rows collection looks like so (discussion is welcomed, it is not the best implementation but idea of adding values row by row in some shape or form is good).

```csharp
  public class SqlInsertValuesCollection : IEnumerable<SqlRow>
  {
    private IReadOnlyList<SqlColumn> columns;
    private List<SqlRow> rows = new();

    public IReadOnlyList<SqlColumn> Columns => columns ?? Array.Empty<SqlColumn>();

    public int RowCount=> rows.Count;

    public void AddRow(Dictionary<SqlColumn, SqlExpression> row)
    {
      if (rows.Count == 0) {
        // save columns order as header for further rows to match;
        columns = row.Keys.ToList();
        rows.Add(SqlDml.Row(row.Values.ToList()));
      }
      else {
        if (columns.Count != row.Count)
          throw new ArgumentException("Inconsistent row lenght");
        if (row.Keys.SequenceEqual(columns)) {
          //fast addition
          rows.Add(SqlDml.Row(row.Values.ToList()));
        }
        else {
          //re-arrange values to be the same order
          //and also make sure all columns exist
          var rowList = new List<SqlExpression>();
          foreach (var column in columns)
            if (row.TryGetValue(column, out var value))
              rowList.Add(value);

          rows.Add(SqlDml.Row(rowList));
        }
      }
    }

    public IEnumerator<SqlRow> GetEnumerator() => rows.GetEnumerator();

    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();

    internal SqlInsertValuesCollection Clone(SqlNodeCloneContext context)
    {
      var clone = new SqlInsertValuesCollection();

      var clonedList = new List<SqlColumn>(columns.Count);
      foreach (var oldColumn in columns)
        clonedList.Add((SqlColumn) context.NodeMapping[oldColumn]);
      clone.columns = clonedList;

      clone.rows = new List<SqlRow>(rows.Count);
      foreach(var oldRow in rows)
        clone.rows.Add((SqlRow) oldRow.Clone());

      return clone;
    }
  }
```

I chose ```SqlRow``` as representation of one row because it is translated into ```(value1, value2, ... ,valueN )```, it was convenient for proof-of-concept, we can store and/or return rows internally in different forms, interfaces which the collection implements are discussable as well as actual implementation.

SergeiPavlov:
Looks good to me

alex-kulakov:
If so, then slightly improved implementation (removing rows, getting rows by index, clear all the rows, ```IReadOnlyList<SqlRow>```, summaries)

```csharp
using System;
using System.Collections;
using System.Collections.Generic;
using System.Linq;
using Xtensive.Core;

namespace Xtensive.Sql.Dml.Collections
{
  /// <summary>
  /// Collection of values of <see cref="SqlInsert"/>.
  /// </summary>
  public sealed class SqlInsertValuesCollection : IReadOnlyList<SqlRow>
  {
    private IReadOnlyList<SqlColumn> columns;
    private List<SqlRow> rows = new();

    /// <summary>
    /// The columns collection has values for.
    /// </summary>
    public IReadOnlyList<SqlColumn> Columns => columns ?? Array.Empty<SqlColumn>();

    /// <summary>
    /// Count of rows.
    /// </summary>
    public int Count => rows.Count;

    /// <summary>
    /// Gets row by index.
    /// </summary>
    /// <param name="index"></param>
    /// <returns></returns>
    public SqlRow this[int index] => rows[index];

    /// <summary>
    /// Adds column-to-value mapped collection of values as row.
    /// </summary>
    /// <param name="row">column-to-value mapped collection of values</param>
    /// <exception cref="ArgumentNullException"><paramref name="row"/> value is null.</exception>
    /// <exception cref="ArgumentException">Count of values between already added rows and <paramref name="row"/> 
    /// -or- particular column in <paramref name="row"/> is not presented in already added rows.</exception>
    public void Add(Dictionary<SqlColumn, SqlExpression> row)
    {
#if NET6_0_OR_GREATER
      ArgumentNullException.ThrowIfNull(row, nameof(row));
#else
      ArgumentValidator.EnsureArgumentNotNull(row, nameof(row));
#endif

      if (rows.Count == 0) {
        // save columns order as header for further rows to match;
        columns = row.Keys.ToList();
        rows.Add(SqlDml.Row(row.Values.ToList()));
      }
      else {
        if (columns.Count != row.Count)
          throw new ArgumentException("Inconsistent row length.");
        if (row.Keys.SequenceEqual(columns)) {
          //fast addition
          rows.Add(SqlDml.Row(row.Values.ToList()));
        }
        else {
          //re-arrange values to be the same order
          //and also make sure all columns exist
          var rowList = new List<SqlExpression>();
          foreach (var column in columns) {
            if (row.TryGetValue(column, out var value)) {
              rowList.Add(value);
            }
            else {
              throw new ArgumentException(string.Format("There is no mentioning of column '{0}' in previously added rows.", column.Name));
            }
          }

          rows.Add(SqlDml.Row(rowList));
        }
      }
    }

    /// <summary>
    /// Removes row by index.
    /// </summary>
    /// <param name="index">The index of row to remove.</param>
    public void RemoveAt(int index)
    {
      rows.RemoveAt(index);
      if (rows.Count == 0) {
        columns = null;
      }
    }

    /// <summary>
    /// Clears rows and columns.
    /// </summary>
    public void Clear()
    {
      rows = new List<SqlRow>();
      columns = null;
    }

    /// <inheritdoc/>
    public IEnumerator<SqlRow> GetEnumerator() => rows.GetEnumerator();

    /// <inheritdoc/>
    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();

    internal SqlInsertValuesCollection Clone(SqlNodeCloneContext context)
    {
      var clone = new SqlInsertValuesCollection();

      if (rows.Count == 0) {
        return clone;
      }

      var clonedList = new List<SqlColumn>(columns.Count);
      foreach (var oldColumn in columns) {
        clonedList.Add((SqlColumn) context.NodeMapping[oldColumn]);
      }
      clone.columns = clonedList;

      clone.rows = new List<SqlRow>(rows.Count);
      foreach(var oldRow in rows) {
        clone.rows.Add((SqlRow) oldRow.Clone());
      }

      return clone;
    }
  }
}
```

SergeiPavlov:
I have added `SqlInsertValuesCollection.cs` :
https://github.com/DataObjects-NET/dataobjects-net/pull/237/commits/a60cd1840877945794b7c0ed332f696944458162

You are free to create sub-PRs with any requestd changes. This would be much easier than copy/paste from comments

alex-kulakov (approved):

PR #242, opened 2022-04-26
Optimize Build Domain operation

SergeiPavlov:
This huge PR is optimization of heavy `Build Domain` operation + some Refactoring.

Benchmarks:

Before:
```
|      Method |    Mean | Error |     Min |     Max |    Gen 0 | Allocated |
|------------ |--------:|------:|--------:|--------:|---------:|----------:|
| BuildDomain | 1.342 s |    NA | 1.342 s | 1.342 s | 250.0000 |    444 MB |
```

After:
```
|      Method |     Mean | Error |      Min |      Max | Allocated |
|------------ |---------:|------:|---------:|---------:|----------:|
| BuildDomain | 994.3 ms |    NA | 994.3 ms | 994.3 ms |    325 MB |
```

I have little hope that it can be merged as whole. But lets discuss which parts of it may be acceptable as separate commits.


alex-kulakov:
It would be nice if there was an explanation for public APIs breaking changes (any public type and/or its members which accessors allow to use them from the outside of the project are considered public APIs); I will ask about them anyway and you will have to revise your huge commit and write answers. Since DataObjects.Net is not an end-user solution but basis for other solutions, we have to change public APIs smoothly by using ```[Obsolete]``` when we want to change such APIs as a general rule, though sometimes we make breaking changes but there must be a serious reason for that.

I would also appreciate a chain of smaller self-explanatory commits with changes that are united by one meaning instead of garbage-binning all the changes into one huge commit, it will also be beneficial for ones who might want to figure something out by the history of changes.

SergeiPavlov:
Added old API for backward compatibility: https://github.com/DataObjects-NET/dataobjects-net/pull/242/commits/1f0ba1d3d8395c9a8691cb66e46d2f3c5cc0d26a

The reason to convert these methods into properties:
1. They return immutable data without any side effect
2. the `recursive` argument is always statically known on caller's side. So it is optimial to split them into 2 methods: recorsive and non-recursive 
3. For more readability.

>I would also appreciate a chain of smaller self-explanatory commits with changes that are united by one meaning
I totally agree with this. The aim of this huge PR: to ask you which parts of it are worth enough to be merged into `master`?
After getting response I will create a separate small commits with that changes.

alex-kulakov:
> I totally agree with this. The aim of this huge PR: to ask you which parts of it are worth enough to be merged into master?
> After getting response I will create separate small commits with those changes.

Ok, then :)  I'll give my feedback of course, but it will take time to review and suggest certain changes structure

alex-kulakov:
Please update the branch from master. I did some changes that should be presented here, for instance, changes of TypeHelper

alex-kulakov:
The TupleDescriptor has some compilation errors - because of brackets and because of ```is null``` check in the Equals(TupleDescriptor) method because of merge, I clearly see that it was OK before .

SergeiPavlov:
> The TupleDescriptor has some compilation errors - because of brackets and because of `is null` check in the Equals(TupleDescriptor) method because of merge, I clearly see that it was OK before .
> 
> ModelBuilder is also merged badly

Lets focus on the subset PR first: https://github.com/DataObjects-NET/dataobjects-net/pull/271
If/after it will be merged I will fix build of this to avoid double merging efforts.

alex-kulakov:
I think TupleDescriptor.Equals() should look like
```csharp

    public bool Equals(TupleDescriptor other)
    {
      if (FieldTypes == null) {
        return other.FieldTypes == null;
      }
      if (other.FieldTypes == null || FieldCount != other.FieldCount) {
        return false;
      }

      for (int i = 0; i < FieldCount; i++) {
        if (FieldTypes[i] != other.FieldTypes[i]) {
          return false;
        }
      }
      return true;
    }
```
otherwise, NRE appear on some comparisons.

There is a problem in AttributeHelper. It should end with following methods
```csharp
    private static IReadOnlyList<Attribute> GetAttributes(MemberInfo member, Type attributeType, AttributeSearchOptions options) =>
      AttributesByMemberInfoAndSearchOptions.GetOrAdd(
        new AttributesKey(member, attributeType, options),
        ExtractAttributes
      );

    private static List<Attribute> GetAttributesAsNewList(this MemberInfo member, Type attributeType)
    {
      var attrObjects = member.GetCustomAttributes(attributeType, false);
      var attrs = new List<Attribute>(attrObjects.Length);
      for (int i = 0, count = attrObjects.Length; i < count; ++i) {
        attrs.Add((Attribute) attrObjects[i]);
      }
      return attrs;
    }

    private static IReadOnlyList<Attribute> ExtractAttributes((MemberInfo member, Type attributeType, AttributeSearchOptions options) t)
    {
      (var member, var attributeType, var options) = t;

      var attributesAsObjects = member.GetCustomAttributes(attributeType, false);
      var attributesCount = attributesAsObjects.Length;

      var attributes = attributesCount > 0
        ? attributesAsObjects.Cast<Attribute>().ToList(attributesCount)
        : null;

      if (options != AttributeSearchOptions.InheritNone) {
        if (attributesCount == 0) {
          if ((options & AttributeSearchOptions.InheritFromPropertyOrEvent) != 0
              && member is MethodInfo m
              && ((MemberInfo) m.GetProperty() ?? m.GetEvent()) is MemberInfo poe) {
            attributes = poe.GetAttributesAsNewList(attributeType);
          }
          if ((options & AttributeSearchOptions.InheritFromBase) != 0
              && (options & AttributeSearchOptions.InheritFromAllBase) == 0
              && member.GetBaseMember() is MemberInfo bm) {
            var attrsToAdd = GetAttributes(bm, attributeType, options);
            if (attrsToAdd.Count > 0) {
              (attributes ??= new List<Attribute>(attrsToAdd.Count)).AddRange(attrsToAdd);
            }
          }
        }

        if ((options & AttributeSearchOptions.InheritFromAllBase) != 0
            && member.DeclaringType != WellKnownTypes.Object
            && member.GetBaseMember() is MemberInfo bm2) {
          var attrsToAdd = GetAttributes(bm2, attributeType, options);
          if (attrsToAdd.Count > 0) {
            (attributes ??= new List<Attribute>(attrsToAdd.Count)).AddRange(attrsToAdd);
          }
        }
      }

      return (IReadOnlyList<Attribute>) attributes ?? Array.Empty<Attribute>();
    }
```

SergeiPavlov:
> therwise, NRE appear on some comparisons.
> 
> There is a problem in AttributeHelper. It should end with following methods

I have made these changes

alex-kulakov:
The last merge has managed to break even more files :)

SergeiPavlov:
Lets close it as obsolete
I have very limited time to dig in this direction

alex-kulakov:
Please, revert last merge, make it be only "Change TupleDescriptor.Equals()" and "Changed AttributeHelper" applied. The branch will be in not broken state. After that I will be able to resolve conflicts on my own.

SergeiPavlov:
> Please, revert last merge, make it be only "Change TupleDescriptor.Equals()" and "Changed AttributeHelper" applied.

Rereverted. but anyway this PR is very huge to continue work on in

alex-kulakov:
I've changed the target branch to a temp one to fix problems if any appear.

PR #248, opened 2022-05-04
Optimizing refactoring: convert EnsureNotLocked() from extension to Method

SergeiPavlov:
I see `EnsureNotLocked()` is a hotspot during profiling.
All invocations of EnsureNotLocked() are for this. We can convert it from extension to method.
This adds following optimizations:

* No need for null check because C# guarantees this is non-null (it can be avoided by manual IL code generating).
* Access to IsLocked propery via interface is much slower than access to it as a field.

Also:
* Fix documentation description of ILockable.Lock() - the only its implementation calls Lock(recursive: true). It is recursive by default.
* Remove `Unlock()` method as non-usecd


Orm/Xtensive.Orm/Core/Extensions/LockableExtensions.cs
```diff
@@ -1,26 +0,0 @@
...
-  /// <see cref="ILockable"/> related extension methods.
-  /// </summary>
-  public static class LockableExtensions
-  {
-    /// <summary>
-    /// Ensures <paramref name="lockable"/> is not locked (see <see cref="ILockable.Lock()"/>) yet.
-    /// </summary>
-    /// <param name="lockable">Lockable object to check.</param>
-    /// <exception cref="InstanceIsLockedException">Specified instance is locked.</exception>
-    public static void EnsureNotLocked(this ILockable lockable)
```
alex-kulakov, line 19: Don't delete it, mark it as Obsolete like so

```
    [Obsolete("Use LockableBase.EnsureNotLocked method instead.")]
    public static void EnsureNotLocked(this ILockable lockable)
    {
      ArgumentValidator.EnsureArgumentNotNull(lockable, "lockable");
      if (lockable.IsLocked)
        throw new InstanceIsLockedException(Strings.ExInstanceIsLocked);
    }
```

Orm/Xtensive.Orm/Core/LockableBase.cs
```diff
@@ -16,35 +16,29 @@ namespace Xtensive.Core
...
-    private bool isLocked;
 
     /// <inheritdoc/>
-    public bool IsLocked
+    public bool IsLocked { [DebuggerStepThrough] get; private set; }
+
+    /// <summary>
+    /// Ensures the object is not locked (see <see cref="ILockable.Lock()"/>) yet.
+    /// </summary>
+    /// <exception cref="InstanceIsLockedException">Specified instance is locked.</exception>
```
alex-kulakov, line 26: We don't specify anything here, like we do with extenson

```suggestion
    /// <exception cref="InstanceIsLockedException">The instance is locked.</exception>
```

Orm/Xtensive.Orm/Core/LockableBase.cs
```diff
@@ -16,35 +16,29 @@ namespace Xtensive.Core
...
+    public void EnsureNotLocked()
     {
-      [DebuggerStepThrough]
-      get { return isLocked; }
+      if (IsLocked) {
+        throw new InstanceIsLockedException(Strings.ExInstanceIsLocked);
+      }
     }
 
+
```
alex-kulakov, line 34: What's this extra line for?

SergeiPavlov: Just an unnoticed insertion. removed it.

Orm/Xtensive.Orm/Core/Extensions/LockableExtensions.cs
SergeiPavlov: done

alex-kulakov (approved):

PR #251, opened 2022-06-06
Add support of DateOnly/TimeOnly types

SergeiPavlov:
Experimental implementation for MSSQL only

Enabled by conditional compilation for .NET6

It can be disabled by setting env var `DO_DATEONLY=false` or by property in `User.Directory.Build.props`


alex-kulakov:
This is great PR! I appreciate it. I'll merge it to a proxy branch where I add support for these times to other providers.

alex-kulakov:
But instead of DO_DATEONLY I would use the built-in NET6 compile time variable, so I will remove this env var mentions from the code, are you OK with it?

SergeiPavlov:
> But instead of DO_DATEONLY I would use the built-in NET6 compile time variable, so I will remove this env var mentions from the code, are you OK with it?

It would be Ok for us but not for other DO users.
`DateOnly` type is supported for SqlServer only

With other DB engine the tests will fail under .NET6 without `DO_DATEONLY` conditional compilation

alex-kulakov:
SQL's ```date``` and ```time``` are supported by majority the majority of RDBMSs. Client libraries can be a problem but we can do the same as we did for ```DateTimeOffset``` - make them unsupported. But I think we will have no problems especially if there is a way to read SQL's ```date``` as ```DateTime``` value and ```time``` as something else and then convert them to DateOnly and TimeOnly respectively.

PR #255, opened 2022-06-22
Fix closure type instance caching problem (issue #224)

alex-kulakov:
Fixes capture of closure type instance declared in #224, reverts previous changes made in #225


alex-kulakov:
@SergeiPavlov , since you're the person who created this check, please take a look.

SergeiPavlov (approved):

PR #256, opened 2022-06-25
Refactor to use C#7 'is null' & C#9 'is not null' operators

SergeiPavlov:
Benchmarking shows that  `is null` operator is ~30% faster than `.IsNullReference()` extension
`object.ReferenceEquals()` has the same performance (it is inlined), but `is null` is more readable

Measured iterations over `object[1000]` array
```
|                       Method |     Mean | Error |      Min |      Max | Allocated |
|----------------------------- |---------:|------:|---------:|---------:|----------:|
|        IsNullReference_Bench | 3.560 us |    NA | 3.560 us | 3.560 us |         - |
| IsNullReferenceInlined_Bench | 2.798 us |    NA | 2.798 us | 2.798 us |         - |
|        ReferenceEquals_Bench | 2.045 us |    NA | 2.045 us | 2.045 us |         - |
|                 IsNull_Bench | 2.000 us |    NA | 2.000 us | 2.000 us |         - |
|              EqualNull_Bench | 2.040 us |    NA | 2.040 us | 2.040 us |         - |                     // ==
```


SergeiPavlov:
Replaced by https://github.com/DataObjects-NET/dataobjects-net/pull/257

PR #259, opened 2022-07-06
DelayedQueryWithDtoCtorParamTest test

SergeiPavlov:


alex-kulakov:
I would throw an exception in such cases instead of not caching query. The query is not cacheable by default. We can mislead user and not cache query and user will go back to us and ask why we don't cache his query, or we can tell the truth that we cannot cache such query and offer user to declare a local variable that will be processed just fine. The third option is providing a way to pass such objects to the query and somehow put such objects to ParameterContext and resolve all the constants that outside of closure to some variable in ParameterContext instance.

PR #264, opened 2022-07-12
Obsolete types and types' members removal

alex-kulakov:


alex-kulakov:
@SergeiPavlov can you check you don't use any of removed things? Note that Session.SelectStorageNode() is also deleted

SergeiPavlov:
> @SergeiPavlov can you check you don't use any of removed things? Note that Session.SelectStorageNode() is also deleted

I confirm, our app does not use any of of this removed stuff.
Checked build with this branch.

PR #265, opened 2022-07-21
Small optimizations (#74)

SergeiPavlov:
* Optimize VisitMethodCall()

* Optimize Provider constructor

* Use Array.Empty<T>() instead of allocating Empty array

* Get rid of unused PriorityQueue<> class

* VisitNullable()


alex-kulakov:
> Get rid of unused PriorityQueue<> class

No public classes deletions allowed, make it [Obsolete]

Orm/Xtensive.Orm/Orm/Rse/Providers/Provider.cs
```diff
@@ -136,7 +127,7 @@ protected virtual string ParametersToString()
     /// </summary>
     /// <param name="type">The type of the provider.</param>
     /// <param name="sources"><see cref="Sources"/> property value.</param>
-    protected Provider(ProviderType type, params Provider[] sources)
+    protected Provider(ProviderType type, IReadOnlyList<Provider> sources)
```
alex-kulakov, line 130: Can you explain what was the problem here? Native array is better, isn't it?
Was it for safety reasons so user code wouldn't be able to change items?

This is a breaking change which generates a breaking change of the ```ExecutableProvider``` ctor, so I need to evaluate its pros.

alex-kulakov (changes requested):

Orm/Xtensive.Orm/Orm/Providers/SqlSelectProcessor.cs
```diff
@@ -562,6 +530,13 @@ public void Visit(SqlExpression sqlExpression)
       sqlExpression.AcceptVisitor(this);
     }
 
+    public void VisitNullable(SqlExpression sqlExpression)
```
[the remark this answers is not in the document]
SergeiPavlov: done

Orm/Xtensive.Orm/Orm/Rse/Providers/Provider.cs
SergeiPavlov: The only reason: make `Provider.Sources` readonly collection

SergeiPavlov:
> > Get rid of unused PriorityQueue<> class
> 
> No public classes deletions allowed, make it [Obsolete]. Removing the test for this collection is OK, though.

`PrioriyQueue` is standard class in .NET6+
https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.priorityqueue-2?view=net-7.0

alex-kulakov (approved):

PR #266, opened 2022-08-02
Build and pack binaries for net6.0 as well

alex-kulakov:
Adds net6.0 to the list of target frameworks. By default build will happen for both ```net6.0``` and ```net5.0``` (to keep build simple ```dotnet build -c Release```). If it is needed to set either net6.0 or net5.0 the environment variable ```DO_TargetFramework``` is still there to override default frameworks list.

Warnings were resolved and some misleading methods were renamed for clarity.

Some obsoletes hashing algorithms remain being in use because the alternative dotnet wants us to use, ```.Create()```, can be very slow. For example, SHA1 results:

|                          Method |       Mean |    Error |   StdDev | Rank |  Gen 0 |  Gen 1 |  Gen 2 | Allocated |
|-------------------------------- |-----------:|---------:|---------:|-----:|-------:|-------:|-------:|----------:|
|                     SHA1ManagedCreate | 2,953.1 ns | 28.13 ns | 24.94 ns |    5 | 0.0877 | 0.0420 |      - |     424 B |
|               SHA1ManagedNew |   544.0 ns | 11.20 ns | 32.50 ns |    1 | 0.0267 | 0.0105 | 0.0010 |     128 B |
|       SHA1CryptoServiceProviderCreate | 2,584.1 ns |  9.89 ns |  9.25 ns |    4 | 0.0267 |      - |      - |     136 B |
| SHA1CryptoServiceProviderNew |   598.5 ns | 15.21 ns | 44.37 ns |    2 | 0.0372 | 0.0181 | 0.0010 |     176 B |
|                            Sha1Create |   817.3 ns | 15.96 ns | 44.22 ns |    3 | 0.1001 | 0.0496 |      - |     472 B |

SHA1ManagedCreate - ```var alg = SHA1.Create("System.Security.Cryptography.SHA1Managed");```
SHA1ManagedNew - ```var alg = new SHA1Managed();```

SHA1CryptoServiceProviderCreate - ```var alg = SHA1.Create("System.Security.Cryptography.SHA1CryptoServiceProvider");```
SHA1CryptoServiceProviderNew - ```var alg = new SHA1CryptoServiceProvider();```

Sha1Create - ```var alg = SHA1.Create("SHA1");```

**UPDATE:** Added build configurations Debug-NET5/Debug-NET6 and Release-NET5/Release-NET6 with target frameworks assigned according to their name endings, the configurations are primarily for development.


alex-kulakov:
@SergeiPavlov Take a look. Multi-targeting is pain in the neck while developing though. I'm thinking of solution that would be convenient. My first intention is to use configurations as a fast way to switch context between net6 to net5 during development, in Directory.Build.props it would be like so:
```xml
  <!-- Debug-NET6, Release-NET6 are mostly for development convenience -->
  <PropertyGroup Condition = "$(Configuration.Contains('NET6')) == 'true'">
    <TargetFrameworks>net6.0</DoTargetFrameworks>
  </PropertyGroup>

  <!-- Debug-NET5, Release-NET5 are mostly for development convenience -->
  <PropertyGroup Condition = "$(Configuration.Contains('NET5')) == 'true'">
    <TargetFrameworks>net5.0</DoTargetFrameworks>
  </PropertyGroup>

  <!--Release and Debug are for final builds, builds on build server, etc. Target frameworks are configurable here -->
  <PropertyGroup Label="Defaults" Condition=="$(Configuration.Contains('NET5')) == 'false' AND $(Configuration.Contains('NET6')) == 'false'">
    <Configuration Condition="$(Configuration) == ''">Debug</Configuration>
    <TargetFrameworks>$(TargetFrameworks)</TargetFrameworks> <!-- the property -->
    <TargetFrameworks Condition="'$(TargetFrameworks)'==''">$(DO_TargetFrameworks)</TargetFrameworks> <!-- env var -->
    <TargetFrameworks Condition="'$(TargetFrameworks)'==''">net6.0;net5.0</TargetFrameworks> <!-- fallback to default -->
  </PropertyGroup>
```

SergeiPavlov:
We tuned multi-targeting 3 years ageo.
As I remember the `TargetFrameworks`/`TargetFrameworks` properties cannot be set dependent on Configuration. 
The onluy way we found to control them: via environment vars

alex-kulakov:
Builds on configurations are building binaries correctly. There are some issues with Nuget packages which I'm going to resolve tomorrow.

alex-kulakov:
Apparently I can change  ```TargetFrameworks``` according to condition but VS handles switches between build configurations poorly - some files in ```obj``` folders are not recreate by VS when configuration is changed so when I change ```TargetFrameworks``` from ```net6.0``` to ```net5.0``` (or vice versa) these files "remember" previous framework and errors.

I'm pretty much sure that there is some target I can use to drop these files on certain point and they will be recreated because ```dotnet build``` works fine.

SergeiPavlov:
> Apparently I can change `TargetFrameworks` according to condition but VS handles switches between build configurations poorly - some files in `obj` folders are not recreate by VS when configuration is changed so when I change `TargetFrameworks` from `net6.0` to `net5.0` (or vice versa) these files "remember" previous framework and errors appear.

I think `IntermediateOutputPath` should include ` $(TargetFramework)
https://github.com/DataObjects-NET/dataobjects-net/pull/266/files#diff-9da24614831c308827a1ae533ffea392c97638c261dd42bd0f5226baa136d16eR75

Else two frameworks share files.

alex-kulakov:
No, it is not, I intentionally removed ```$(TargetFramework)``` from ```IntermediateOutputPath``` because target framework is added to the end of this path during build. After that we had something like ```_Build\obj\Xtensive.Orm\Debug\net5.0\net5.0\``` and ```_Build\obj\Xtensive.Orm\Debug\net6.0\net6.0\```. I noticed this in build log and then removed extra ```$(TargetFramework)```. The only exception is Weaver project but it has ```AppendTargetFrameworkToOutputPath``` set to false, presumably for some purpose (and I have no time to check why :) ) , so I added ```$(TargetFramework)``` within Weaver project manually.

The problem, as I understand, is in ```ProjectAssetsFile```, at least visual studio mentions it in errors when make a switch. It has a ```targets``` collection and when configuration switches from net6.0 to net5.0 then VS tell that there is no net5.0 and obviously there is only net6.0 in the ```targets``` collection. I'm not sure whether I can put them to a ```$(TargetFramework)``` subfolder. I try it though

alex-kulakov:
It finally works! Works in VS, I'm not sure about Rider or other IDEs. According to build log Rider also sets ```$(BuildingInsideVisualStudio)``` to true so restore of packages after changing configuration should work too.  I added a property that allow to add some checks to identify some other IDEs if they have such unique properties.

@SergeiPavlov , check this out, I don't know your teams' IDE preferences so it would be nice to check on all variety of used IDEs (or we could add them on the go). Now It should be much easier than setting an environment variable.

BTW. It is required to make ```git clean -xdf``` because structure of ```obj``` folder content is changed .

SergeiPavlov:
> @SergeiPavlov , check this out, I don't know your teams' IDE preferences so it would be nice to check on all variety of used IDEs (or we could add them on the go). Now It should be much easier than setting an environment variable.

I confirm. Either kind of build I use does work:
* dotnet build 
* VS 2022

Issue #81, opened 2020-10-21
DirectPersistentAccessor.SetReferenceKey change Tuple but does not register entity as modified


alex-kulakov:
This happens only with saved entities (ones with PersistanceState equals PersistanceState.Synchronized). New entities are already registered for following persist so any change of them will be saved anyway.

Test to reproduce the problem
```
    [Test]
    public void SetReferenceFieldOfExistingEntityTest()
    {
      var entityId = 0;
      var entityRefId = 0;
      using (var session = Domain.OpenSession())
      using (var t = session.OpenTransaction()) {
        var accessor = session.Services.Get<DirectPersistentAccessor>();
        var myEntity = (MyEntity) accessor.CreateEntity(typeof (MyEntity));
        entityId = myEntity.Id;
        entityRefId = new MyRefEntity().Id;
        t.Complete();
      }

      using (var session = Domain.OpenSession())
      using (var t = session.OpenTransaction()) {
        var myEntity = session.Query.All<MyEntity>().First(e => e.Id==entityId);
        var referenceEntity = session.Query.All<MyRefEntity>().First(e => e.Id==entityRefId);

        var field = myEntity.TypeInfo.Fields["Reference"];
        var accessor = session.Services.Get<DirectPersistentAccessor>();
        accessor.SetReferenceKey(myEntity, field, referenceEntity.Key);
        Assert.That(myEntity.PersistenceState, Is.EqualTo(PersistenceState.Modified));
        t.Complete();
      }

      using (var session = Domain.OpenSession())
      using (var t = session.OpenTransaction()) {
        var myEntity = session.Query.All<MyEntity>().First(e => e.Id==entityId);
        Assert.That(myEntity.Reference, Is.Not.Null);
        Assert.That(myEntity.Reference.Id, Is.EqualTo(entityRefId));
      }
    }
```

PR #270, opened 2022-08-11
On Node removing remove related keys from QueryCache (#81)

SergeiPavlov:
This bug appears in following scenareo:

1. Create some cached query with Node "foo"
2. call `Domain.StorageNodeManager.RemoveNode()`
3. rename schema of the Node
4. Add node with the same name "foo" but other schema
5. Invoke the same query which was cached in step 1. - it references the old schema


alex-kulakov:
Hello,

Can you reproduce this on current master?

Can you open an issue for this with small [Test] method which shows the problem? This should be applicable to 6.0 and 7.0 too so it should be resolved in 6.0. As I understand this is not a regular case, personally I would expect nodes to be with same settings. If I'm right then I would offer a StorageNodeManager.Remove(string nodeID, bool clearQueryCache = false) method.

alex-kulakov:
When you offered a united cache for all nodes, I thought about having nodes which would store their portion of QueryCache so instead of one giant cache we would have a number of smaller caches. This would allow to release resources on node removal but in this case all re-added nodes would start working with cold query cache which is drawback.

SergeiPavlov:
> Hello @SergeiPavlov,
> 
> Can you reproduce this on current master?
> 
> Can you open an issue for this with small [Test] method which shows the problem? This should be applicable to 6.0 and 7.0 too so it should be resolved in 6.0. As I understand this is not a regular case, personally I would expect nodes to be with same settings. If I'm right then I would offer a StorageNodeManager.Remove(string nodeID, bool clearQueryCache = false) method.

I hardly have to write DO tests. The case was detected in real app and fixed by this code.

alex-kulakov:
Well, I'm pretty sure that I understood the problem, but would be nice to have real example to exclude any misunderstanding, I'll try to reproduce it. How often such scenario happening? Was I right about rareness of this case?

alex-kulakov:
I made some changes for this issue for 6.0 branch since it is affected too (PR #296), the changes contain test cases, you can take a look. After merge I'll merge it to 7.0 and then to master with required changes.

SergeiPavlov:
> Well, I'm pretty sure that I understood the problem, but would be nice to have real example to exclude any misunderstanding, I'll try to reproduce it. How often does such scenario happen? Was I right about rareness of this case?

The scenario is not typical. This is the steps:
1. We rename schema for tenant, actually we move tenant DB to another server and db
2. then we update NodeRecord
3.  start getting error like:
```
SQL error occured.SQL error details 'Type: SyntaxError;'Query 'SELECT ... FROM [spraleigh].[User] ...]'Original message 'Invalid object name 'oldschema.User'.'
```
due to using old schema name

alex-kulakov:
Let me put it this way.
1) How frequently does your app call Domain.Build()? every day, every week, every month?
2) How frequently does your app add nodes to Domain and remove nodes from Domain? daily? weekly?
3) How frequently do you move tenants between schemas/DBs/servers/ whatever?

SergeiPavlov:
> Let me put it this way.
> 
> 1. How frequently does your app call Domain.Build()? every day, every week, every month?
> 2. How frequently does your app add nodes to Domain and remove nodes from Domain? daily? weekly?
> 3. How frequently do you move tenants between schemas/DBs/servers/ whatever?

1. Hundreeds times per day
2. In normal mode it only adds thosands times per one Domain
3. It is manual operation. ~100 times per week

alex-kulakov:
@SergeiPavlov, the changes from 6.0 with StorageNodeManager.Remove(string nodeId, bool clearQueryCache = false) option were merged to 7.0 and then to master (with required changes for newer code). I can close this PR without merge, right?

SergeiPavlov:
> @SergeiPavlov, the changes from 6.0 with StorageNodeManager.Remove(string nodeId, bool clearQueryCache = false) option were merged to 7.0 and then to master (with required changes for newer code). I can close this PR without merge, right?

Yes. I'm cloning it

SergeiPavlov:
Replaced by https://github.com/DataObjects-NET/dataobjects-net/commit/bb21e6f31531a0df9bc8a1c2d4417329cd65623c

Issue #84, opened 2020-10-30
Invalid cast exception in the scalar query with a type cast


AlexUstinov:
In version 7.0 the following code fails
```csharp
using System;
using System.Linq;
using System.Reflection;
using NUnit.Framework;
using Xtensive.Orm.Configuration;
using Xtensive.Orm.Model;
using Xtensive.Orm.Tests.Issues.GithubIssue_xxx_InvalidCastModel;

namespace Xtensive.Orm.Tests.Issues.GithubIssue_xxx_InvalidCastModel
{
  [HierarchyRoot(InheritanceSchema.ConcreteTable)]
  public class Item : Entity
  {
    public Item(Session session)
      : base(session)
    { }

    [Field, Key]
    public long Id { get; private set; }

    [Field]
    public int IntField { get; set; }
  }
}

namespace Xtensive.Orm.Tests.Issues
{
  [TestFixture]
  public class GitHubIssue_xxx_InvalidCast : AutoBuildTest
  {
    protected override DomainConfiguration BuildConfiguration()
    {
      var config = base.BuildConfiguration();
      config.Types.Register(Assembly.GetExecutingAssembly(), typeof (Item).Namespace);
      return config;
    }

    [Test]
    public void InvalidCastOnMaterialization()
    {
      using (var session = Domain.OpenSession()) {
        using (session.OpenTransaction()) {
          _ = new Item(session);
          Assert.AreEqual((int?) 1, session.Query
            .All<Item>()
            .GroupBy(s => s.IntField)
            .Max(g => (int?)g.Count()));
        }
      }
    }
  }
}
```

alex-kulakov:
Hello, the case works fine on 6.0.4. Did it work on 6.0.3?

alex-kulakov:
I have checked the case on 6.0.3 - it fails. After merging 6.0 branch to master it will be fixed on DO 7 as well

alex-kulakov:
I merged 6.0.4 changes to the master branch, ran the case and it case works. The issue is closed.

PR #261, opened 2022-07-08
Use C#9 Covariant return types feature for Clone() implementation for type safety and reduce number of casts

SergeiPavlov:

PR #274, opened 2022-08-22
Avoid unnecessary CompilableProvier castings (#84)

SergeiPavlov:


alex-kulakov:
Hello, so this is similar to the case in #261, right?

SergeiPavlov:
> Hello, so this is similar to the case in #261, right?

Not exaclty.
The aim of #265 was to remove common part of each Clone() method for simplification.

This PR avoids explicit type casts

alex-kulakov:
@SergeiPavlov,

With these changes of ```ProviderVisitor```, it is basically the same as ```CompilableProviderVisitor```. No references to ```ProviderVisitor``` exist, except for usage of it as base class for ```CompilableProviderVisitor```. This means there is no way to have it as direct base class for something in user code. I'm thinking of merging ```ProviderVistior``` and ```CompilableProviderVisitor```.

By the way, some members of ```CompilableProviderVisitor```, for instance, ```VistitTag```, ```VisitDistinct``` methods have specific return types instead of general ```CompilableProvider``` type. Why? To have advantage of covariant returns we need to call method and work with result directly, without any casts, but I can't see any direct usage of results, moreover, they are casted to ```CompilableProvider``` in the ```Visit``` method, so cast to base class happens anyway, right?

SergeiPavlov:
Right now returning `TagProvider` form `VisitTag(TagProvider provider)` is unnecessary, but may be useful in future.
Generally returning subtype is more safe.

Casting to base class is compile-time, so it is free.

alex-kulakov:
But it blocks overriding methods from being able to return something other than provider type of parameter. ```CompilableProviderVisitor``` is used in rewriters which may transform tree of providers including change of type of returning instance. If we changed return type of virtual method this opportunity would be closed. It also may be useful for future, and in case of such need of rewriting we would have to change virtual method signature back to base type - ```CompilableProvider```. Not very good for base class. I assume that overriding methods prevented you from declaring all the virtual ```VisitXxx``` methods to have exact return types, right?

I would vote for keeping ```CompilableProvider``` as return type of virtual methods. I did some benchmarks on dummy providers tree of 200 nodes (only visiting of the tree). Fastest option changes from one to another in some runs but both options are faster than returning ```Provider```.

|                    Method |     Mean |     Error |    StdDev | Rank |  Gen 0 |  Gen 1 | Allocated |
|-------------------------- |---------:|----------:|----------:|-----:|-------:|-------:|----------:|
| ReturnCompilableProviders | 3.106 us | 0.0110 us | 0.0092 us |    1 | 1.6899 | 0.0229 |      8 KB |
|   RetrunExactProviders | 3.139 us | 0.0205 us | 0.0191 us |    1 | 1.6899 | 0.0229 |      8 KB |
|          RerturnProviders | 3.367 us | 0.0062 us | 0.0048 us |    2 | 1.6899 | 0.0229 |      8 KB |

|                    Method |     Mean |     Error |    StdDev | Rank |  Gen 0 |  Gen 1 | Allocated |
|-------------------------- |---------:|----------:|----------:|-----:|-------:|-------:|----------:|
|   RetrunExactProviders | 3.188 us | 0.0302 us | 0.0268 us |    1 | 1.6899 | 0.0229 |      8 KB |
| ReturnCompilableProviders | 3.272 us | 0.0245 us | 0.0229 us |    2 | 1.6899 | 0.0229 |      8 KB |
|          RerturnProviders | 3.501 us | 0.0216 us | 0.0191 us |    3 | 1.6899 | 0.0229 |      8 KB |

I believe that tells that returning ```CompilableProvider``` and having it across the board is good idea overall.

Also, I'm getting rid of ```ProviderVisitor``` as base class of visitors and will make a PR to your branch a bit later today.

SergeiPavlov:
Ok, I've changed return type to `CompilableProvider`

alex-kulakov:
> My guess was that every VisitFoo() method returns the same type as its argument, because it is a kind of filter.

This is true but mostly for sealed classes, also might be true for some in-between classes. For instance, ColumnMappingInspector does not change structure of ```Provider```s tree, but some others may change the tree, optimize it or shrink it. This is why I vote for methods returning ```CompilableProvider`` in ```ComplilableProviderVisitor```, I don't vote for such things in its accessors.

This is what I did in my changes to this PR, I agreed with you and made ancestors take advantage of covariant returns.
The PR I did https://github.com/servicetitan/dataobjects-net/pull/134

alex-kulakov (approved):

PR #275, opened 2022-08-27
Refactor ProviderVisitor to generic form

SergeiPavlov:
* Remove duplicated `switch statement` and convert it to "switch expression"
* Remove unnecessary casts to `CompilableProvider`


alex-kulakov (approved):
Everything is OK but conflicts should be resolved

alex-kulakov (approved):

PR #296, opened 2022-12-28
On Node removing remove related keys from QueryCache if required

alex-kulakov:
Adds extra parameter for StorageNodeManager.Remove() method for being able to remove query cache items dedicated to the removing node


SergeiPavlov:
Looks good to me

PR #301, opened 2023-01-11
Intern short (1..2 chars) contents of TextNode; Statically preallocate Comma TextNode as mostly used

SergeiPavlov:
Most typical TextNode is  `comma+space`
Cached queries may contain Megabytes of them:
image


alex-kulakov:
I'm going to change target branch to the 7.1 one, it's the version where reworked ContainerNode was introduced sounds reasonable to improve it there

alex-kulakov (changes requested):
Static TextNode.Create(string ) methods, as all static methods should be before any instance methods.
So the order should be 
1) public readonly string Test;
2) public static TextNode Create(string text) ...
3) internal override void AcceptVisitor(NodeVisitor vistor)

SergeiPavlov:
> Static TextNode.Create(string ) methods, as all static methods should be before any instance methods.

done

alex-kulakov (approved):

PR #302, opened 2023-01-14
Save 8/4 bytes per node instance

SergeiPavlov:
Potentially is saves megabytes

image

### DIFF
--- a/Orm/Xtensive.Orm/Sql/Compiler/Internals/Nodes/ContainerNode.cs
+++ b/Orm/Xtensive.Orm/Sql/Compiler/Internals/Nodes/ContainerNode.cs
@@ -153,7 +153,7 @@ namespace Xtensive.Sql.Compiler
     {
       FlushBuffer();
       children.Add(node);
-      lastNodeIsText = node.IsTextNode;
+      lastNodeIsText = node is TextNode;
     }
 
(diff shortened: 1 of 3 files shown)


alex-kulakov:
It is 29% slower operation, according to my benchmark runs. You hurt performance for memory. Do you struggle for memory that much?

SergeiPavlov:
> It is 29% slower operation, according to my benchmark runs.

Does this measuremetn include GC-time?
My experience says that GC is one of main factors of performances.
So I have a bias to save more memory at the cost of CPU cycles

alex-kulakov:
> Does this measuremetn include GC-time?

No, it doesn't, I wanted to know what is pure operations cost.

> So I have a bias to save more memory at the cost of CPU cycles.

Not always, we know it :) 

OK, lets not have the field.

Since a Node is created less frequent than checked for its ```TextNode```ness during conversion to command text :) I wanted to cache this info to improve translation because it happens a lot of times for cached queries, it reduces translation overhead for a bit for the cost of size of this field. This was my thoughts, to me faster queries are important, little bit here, little bit there and we have some significant value :)

If this is ok for you highly loaded app then I agree.

alex-kulakov (dismissed):

alex-kulakov (approved):

PR #304, opened 2023-02-27
Upgrade to SqlClient 5.1.0 with DateOnly/TimeOnly support

SergeiPavlov:
SqlClient 5.1.0 has support of `DateOnly/TimeOnly` types
No need for intermediate conversion


alex-kulakov:
If you update a storage-specific thing, don't change base implementation, change and override the provider, changing the base one can break and probably will break other providers.

alex-kulakov:
I'm going to publish an RC version of 7.1 with DateOnly/TimeOnly support across all supported types of storages (PR #307). This PR fits it too. But with exception of these two no merges will happen before RC.

SergeiPavlov:
> If you update a storage-specific thing, don't change base implementation, change and override in the the provider, changing the base one can break and probably will break other providers.

Fixed

alex-kulakov:
Also, for the future, ```TypeMapper``` has clear structure of grouped methods, It would be appreciated if ReadXxx, BindXxx and MapXxx methods were added in their groups. It would smooth review process and give you extra points of karma 😉

alex-kulakov (approved):

PR #318, opened 2023-03-23
Memoize InnerGetTypeName()

SergeiPavlov:
We see in tracing this is a hotspot

### DIFF
--- a/Orm/Xtensive.Orm/Reflection/TypeHelper.cs
+++ b/Orm/Xtensive.Orm/Reflection/TypeHelper.cs
@@ -85,7 +85,7 @@ namespace Xtensive.Reflection
       key.genericDefinition.MakeGenericType(key.typeArgument1, key.typeArgument2);
 
     private static readonly ConcurrentDictionary<(Type Type, bool UseShortForm), string> MemoizedTypeNames = new();
-    
+
     private static int createDummyTypeNumber = 0;
     private static AssemblyBuilder assemblyBuilder;
     private static ModuleBuilder moduleBuilder;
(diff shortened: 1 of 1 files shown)


SergeiPavlov:
I don't remember.
By be just to make the code optimal for any case.

alex-kulakov:
I think I know when you get hot paths for ```GetShortName()```. Usage during Domain build corelates with number of model nodes (```TypeInfo```, ```FieldInfo```, ```IndexInfo``` and ```TypeDef````, ```FieldDef``` and so on). On ```HugeModeUpgrade.RegularModel``` I have 615 calls of ```GetShortName``` during domain build. Real problem is queries, especially ```TranslatorContext.GetApplyParameter()```. During validation of populated data, tests execute queries and the queries make 2600 additional calls of ```GetShortName```.

In various cases we can replace ```GetShortName``` with System.Type.Name, it is valid replacement for types which are guaranteed to be not Nested and not GenericType. Providers are among them. I have found more cases when we can make such replacement. 

I see that System.RuntimeType has cache for names (according to the sources)

```csharp
        public override String Name 
        {
            get 
            {
                return GetCachedName(TypeNameKind.Name);
            }
        }
```

Implicit usage of this cache will help us to not bloat our caches.

Maybe we can start with something like this (#333) and see how it goes. What do you think? I don't know whether you can check results of the PR I mentioned and tell me the outcome.

SergeiPavlov:
I don't mind these changes, but to test them  with our app requires big efforts to create special branches and build.
So I'm going to wait until they will be merged into main branch

alex-kulakov:
Sure, as I thought, it is understandable.

alex-kulakov:
I merged my PR, I postpone this one for results of my PR, whether it did something or not.

PR #326, opened 2023-06-13
Sub query tags

Mee7891:
Found a case when adding a tag into sub query leads to wrong compiled sql


alex-kulakov:
There are a lot of files which don't exist in original repo and you are not allowed to add them. Also unclear how exactly the issue looks like and what you are going to fix and how.

Fix should be with exact changes of the files which contribute to the issue.

Rejected.

Issue #129, opened 2021-04-14
Key field can't be of 'Byte[]' type


matthewhancock:
Hi, I was prototyping with an Entity that has a `byte[]` (converted `Ulid`) as a key. Would it be possible to narrow this filter to allow `byte[]` as a key, or are there other reasons for the limitation?

https://github.com/DataObjects-NET/dataobjects-net/blob/1ca7dfc9a6fb8a996f5b7ed83ebc5680ed6e706a/Orm/Xtensive.Orm/Orm/Building/Validator.cs#L132

Thanks!

alex-kulakov:
Hello Matthew,

From what I know and see I can assume it was made to generalize primary key types across supported RDBMS. (Maybe there is some reasons I don't know yet.)

Some of the supported RDBMS do not allow native equivalent of ```byte[]``` to be part of primary key. Because of that I cannot remove this check completely.

I have found that ULID is compatible with UUID [(here)](https://github.com/ulid/spec). So it is compatible with GUID. If it is then have you tried to create a GUID value from it? Any problems with this I don't know about?

matthewhancock:
The bit ordering and comparison between Guids and binary(16) in SQL server appears to be slightly different: https://github.com/rmalayter/ulid-mssql

> SQL Server sorts UUIDs in an unexpected way; the last 48 bits are sorted first as a big-endian binary unit, followed by other components bits as little-endian chunks.

The first 6 bytes in Ulids are the most critical as they are they are an encoded timestamp which maintains the chronology (the last 10 bytes are random).

I haven't played around enough to know how big an issue it is, but am just a little wary of mapping the two types together because they contain the same amount of bytes. It may also lead to unintentional use or reliance on Guid functionality (e.g. newid()) that would undermine some of the benefits of Ulids.


Also, sorry this is slightly unrelated, it would need to be a byte[] or Guid defined as a field (and not Ulid) because DO doesn't have value converter functionality, like Entity Framework, where you can globally specify a type to how it should be handled by the database, right?

alex-kulakov:
Thank you for the link to performance tests. ULIDs seem to perform better than UUIDs

> Also, sorry this is slightly unrelated, it would need to be a byte[] or Guid defined as a field (and not Ulid) because DO doesn't have value converter functionality, like Entity Framework, where you can globally specify a type to how it should be handled by the database, right?

[...]

PR #327, opened 2023-07-05
`Transaction.Timeout` property (#129)

SergeiPavlov:
Also:
* Refactor `Transaction` class

### DIFF
--- a/Orm/Xtensive.Orm/Orm/Providers/StorageDriver.Operations.cs
+++ b/Orm/Xtensive.Orm/Orm/Providers/StorageDriver.Operations.cs
@@ -397,14 +397,14 @@ namespace Xtensive.Orm.Providers
     #region Sync Execute methods
 
     public int ExecuteNonQuery(Session session, DbCommand command) =>
-      ExecuteCommand(session, command, CommandBehavior.Default, (c, cb) => c.ExecuteNonQuery());
+      ExecuteCommand(session, command, CommandBehavior.Default, static (c, cb) => c.ExecuteNonQuery());
 
     public object ExecuteScalar(Session session, DbCommand command) =>
-      ExecuteCommand(session, command, CommandBehavior.Default, (c, cb) => c.ExecuteScalar());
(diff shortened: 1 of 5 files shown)


alex-kulakov:
1) There is no such native setting as Transaction timeout. It starts and ends when user.
2) Commands are expected to have either default timeout or the timeout which was set in ```SessionConfiguration.CommandTimeout```. But this behavior treats commands unequally which makes them order-dependent. This inequality makes commands execution untransparent and illogical for the user. Depending on this setting the same query may be executed successfully or failed because some artificial obstacle (because otherwise query would've executed successfully).
3) It interferes with async tokens, which adds even more unpredictability to the equation.
4) This custom behavior is possible to realize without changes on DataObjects.Net code. You have events for transaction opening to get transaction, you have extensions collection where you can store whatever you want and you have event on command executing which gives you DbCommand to set ```CommandTimeout``` property.

SergeiPavlov:
> 4\. This custom behavior is possible to realize without changes

Yes, "On events" was our initial implementaion 

But after we implemented another feaure "SessionGuard" for detecting concurrent usage of Session, it had a conflict with `Transaction` feature, because of order of events.
So the Eventless implementation is necessary for us.

alex-kulakov:
It still conflicts with bunch of other cases. I hardly can allow this. It is very shady in terms of predictability. Maybe we can improve extendibility in a way that will allow you to implement it like a session service or something.

What about sessionhandler replacement?

alex-kulakov:
As I understand you need some code which will work not in context of all sessions, like events do, but in context of one exact session. Am I right?

SergeiPavlov:
I did not use SessionHandler, have to learn it.

Regarding to Transaction.Timeout, we need to limit execution time of code like:
```
var tx = session.OpenTransaction();
tx.Timeout = Timespan.FromSeconds(10);
for (int i = 0; i < N; ++i) {
    // some DO query
}
tx.Complete();
```

PR #331, opened 2023-08-10
Improves memory efficiency of building indexes.

alex-kulakov:
- allows to skip some copying of underlying indexes for Union and Join indexes
- shrinks size of final ```ColumnIndexMap``` for real-world models, reduce amount of lists (and capacity changes)

This should be merged after #297.

### DIFF
--- a/Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
+++ b/Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
@@ -399,12 +399,6 @@ namespace Xtensive.Orm.Model
 
     // Constructors
 
-    private IndexInfo()
-    {
-      includedColumns = new ColumnInfoCollection(this, "IncludedColumns");
-      valueColumns = new ColumnInfoCollection(this, "ValueColumns");
-    }
-
     /// <summary>
     /// Initializes a new instance of this class.
     /// </summary>
@@ -494,6 +488,12 @@ namespace Xtensive.Orm.Model
       if (UnderlyingIndexes.Count == 0) {
         throw new ArgumentException(Strings.ExSequenceContainsNoElements, nameof(baseIndexes));
       }
+    }
+
+    private IndexInfo()
+    {
+      includedColumns = new ColumnInfoCollection(this, "IncludedColumns");
+      valueColumns = new ColumnInfoCollection(this, "ValueColumns");
     }
   }
 }
--- a/Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
+++ b/Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
@@ -329,23 +329,33 @@ namespace Xtensive.Orm.Model
       if (!IsPrimary)
         return;
 
-      var system = new List<int>();
+      var keyColumnsCount = keyColumns.Count;
+      var system = new int[keyColumnsCount + 2];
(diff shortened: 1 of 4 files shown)


Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
```diff
@@ -320,32 +320,41 @@ public override void UpdateState()
...
+      var system = new int[keyColumnsCount + 1];
       var lazy = new List<int>();
-      var regular = new List<int>();
+      var regular = new int[Columns.Count - keyColumnsCount];
+
+      var regularIndex = 0;
+      var systemIndex = 0;
 
-      for (int i = 0, count = Columns.Count; i < count; i++) {
+      for (var i = 0; i < Columns.Count; i++) {
```
SergeiPavlov, line 341: Invoking `.Count` on every iteration would be inefficient

Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
```diff
@@ -320,32 +320,41 @@ public override void UpdateState()
...
-            lazy.Add(i);
-          else
-            regular.Add(i);
+          regular[regularIndex++] = i;
         }
       }
 
-      ColumnIndexMap = new ColumnIndexMap(system, regular, lazy);
+      ColumnIndexMap = new ColumnIndexMap(
+        (systemIndex == system.Length) ? system : new ArraySegment<int>(system, 0, systemIndex),
```
SergeiPavlov, line 355: No need for `?:` here
`ArraySegment<int>(system, 0, systemIndex)` works for  `systemIndex == system.Length` case too

`system` casts to `ArraySegment` anyway.

SergeiPavlov:
Looks good to me

Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
```diff
@@ -320,32 +320,41 @@ public override void UpdateState()
...
-          if (item.IsLazyLoad)
-            lazy.Add(i);
-          else
-            regular.Add(i);
+          regular[regularIndex++] = i;
         }
       }
 
-      ColumnIndexMap = new ColumnIndexMap(system, regular, lazy);
+      ColumnIndexMap = new ColumnIndexMap(
```
SergeiPavlov, line 354: We have implicit boxing hire, as `SegmentArray` is value type, and casting value type to interface is boxing always.

So `List<>` may be better. It has not implicit casting overhead and the same allocation overhead

Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
alex-kulakov: I wanted to not have struct wrapper when there is no need for it.

SergeiPavlov: Then you have to use Invoke `new ColumnIndexMap()` in every branch
Type of `(systemIndex == system.Length) ? system : new ArraySegment<int>(system, 0, systemIndex)` subexpression is `ArraySegment`

Orm/Xtensive.Orm/Orm/Model/IndexInfo.cs
alex-kulakov: done

PR #364, opened 2023-12-20
IAsyncEnumerable on DelayedQuery

snaumenko-st:
Introduce IAsyncEnumerable on DelayedQuery to utilize async Session query while asynchronously prefetching data.
We need IAsyncEnumerable implementation here as if it is used inside a Prefetch, which is a fairly common case, it is used as the IEnumerable source and will fall back to a synchronous implementation even in a chain of asynchronous calls. The new unit test shows that behavior.

### DIFF
--- a/Orm/Xtensive.Orm.Manual/Prefetch/PrefetchTest.cs
+++ b/Orm/Xtensive.Orm.Manual/Prefetch/PrefetchTest.cs
@@ -5,11 +5,14 @@
 // Created:    2009.12.24
 
 using System;
+using System.Collections.Generic;
 using NUnit.Framework;
 using System.Linq;
+using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
-using Xtensive.Core;
-using Xtensive.Orm.Internals.Prefetch;
+using Xtensive.Orm.Internals;
+using Xtensive.Orm.Providers;
 using Xtensive.Orm.Services;
 using Xtensive.Orm.Tests;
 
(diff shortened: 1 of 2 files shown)


alex-kulakov:
Hello @snaumenko-st,

Oh my gosh, I can feel your frustration when you had an API to change session handler but no way to get current session handler to instantiate yours. Sorry for that. We'll fix it and provide proper way of changing handlers once we have this API.

I also appreciate the test you have created. I have some comments on it.
1) **Managing session services.** Don't instantiate session services, don't do it in DataObjects.Net project and in yours as well, use ```Session.Services.Get<TServiceType>()``` API, let services be under DataObjects.Net control, it creates them and it should dispose them if needed.
2) **Test.** I need you to move your test case to our main test project - ```Xtensive.Orm.Tests```. It has a group of tests connected to Prefetch functionality, they are placed in Xtensive.Orm.Tests/Storage/Prefetch folder, I think you need to put your test cases into PrefetchTest.cs file. The test class populates database with some data so you just need to make some queries.
Split sync and async cases which are now in one test, let's call tests ```PrefetchOnDelayedQueryTest``` and ```PrefetchOnDelayedQueryAsyncTest```. Please, **append them to the end** of tests region.
Put your session handler to a separate file in Xtensive.Orm.Tests/Storage/Prefetch folder, it may be useful not only in your case, if so, I will move it later to Tests.Framework project to be available across the test projects.

snaumenko-st:
@alex-kulakov Thanks for your comments. I tried fixing the tests as you suggested and they look a little cleaner now.

Issue #394, opened 2024-08-02
EntitySet and ToListAsync extension method runtime error


axmty:
Hi,

In my company we are migrating to DataObjects v7, but we encountered some problems concerning the AsAsync extension method that seemed to disappear, and that we used everywhere (on IQueryable<T> or EntitySet<T> objects).
It seems that we can use ToListAsync instead, but we have a problem when using it on EntitySet.

When reading the code of EntitySet, we expected ToListAsync to work since EntitySet<T> implements IQueryable<T>, but there is a runtime error because EntitySet does not implement IAsyncEnumerable.
Is there a reason why EntitySet<T> does not implement IAsyncEnumerable<T>?

After reading the implementation of Queryable<T>.GetAsyncEnumerator, I think we could do the same in the EntitySet<T> class (that is using EntitySet.Provider to get the same result):

```
public async IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
{
  var result = await Provider.ExecuteSequenceAsync<T>(expression, cancellationToken).ConfigureAwait(false);
  var asyncSource = result.AsAsyncEnumerable().WithCancellation(cancellationToken).ConfigureAwait(false);
  await foreach (var element in asyncSource) {
    yield return element;
  }
}
```

But maybe we are missing something that prevents us to do so.

We found a small workaround (myEntitySet.Where(x => 1 == 1).ToListAsync()) for the moment, but I would greatly appreciate some explanation.

Thanks a lot,
Alex

axmty:
I have opened a PR here : #395 

Let me know if that is ok.

alex-kulakov:
Hello @axmty

I appreciate your help, this is really cool that you want to commit to the project.

But there is a catch when working with/changing code of ```EntitySet<T>``` - it has internal state, a mixture of items that are in storage, locally added items and locally removed items. It was done so to minimize interactions with database, otherwise, changes of collection would spam database with INSERTs/DELETEs. Local state makes it possible to save changes of collection in batches, not wasting time on interactions with database.

[...]

PR #395, opened 2024-08-02
Make EntitySet<T> implement IAsyncEnumerable<T> so that we can use ToListAsync() on EntitySet<T>

axmty:
#394 

In previous version of DataObjects, we could use AsAsync() on both IQueryable<T> and EntitySet<T>.
Now that AsAsync() has been removed, we expect ToListAsync() to work for both IQueryable<T> and EntitySet<T> types, but there is a runtime error since EntitySet<T> does not implement IAsyncEnumerable<T>.

### DIFF
--- a/Orm/Xtensive.Orm.Tests/Linq/EntitySetTest.cs
+++ b/Orm/Xtensive.Orm.Tests/Linq/EntitySetTest.cs
@@ -6,6 +6,7 @@
 
 using System;
 using System.Linq;
+using System.Threading.Tasks;
 using NUnit.Framework;
 using Xtensive.Orm.Tests;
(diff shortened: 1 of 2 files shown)


alex-kulakov:
I'm rejecting this PR. The idea is good, but many things haven't implemented to ```EntitySet```s be fully functional, the collection uses many APIs which also have to support async operations and it is hard to implement from someone outside the team.

Since root idea was to be able to use certain extensions with ```EntitySet```s, I just improved compatibility of those extensions with ```EntitySet```s (PR #413), which is way simpler than implementing full support of IAsyncEnumerable by ```EntitySet``` itself.

PR #468, opened 2026-04-21
Tuple transform refactoring

alex-kulakov:
Heavy refactoring of tuple transformation and ```Rse.Providers``` made by @AlexUstinov

- Many types from ```Xtensive.Tuples.Transform``` namespace were removed, renamed or turned into internal, whole concept simplified and now contains only what's in use
- ```Xtensive.Collections.FixedList3``` is removed and ```Xtensive.Tuples.TupleExtensions```'s methods that used it also removed
- ```Xtensive.Orm.Rse.Providers``` no longer rely on virtual method to build their headers but rather use static methods within type to do so, which reduces overhead significantly by just faster calls.
- Removed some double-checks of incoming parameters or made them a bit faster by using dotnet validators instead of ours ```ArgumentValidator```.
- other slight improvements here and there


AlexUstinov:
😲

